PR #1, opened 2017-06-15
Fix: LT-18378 Update Flex to use the 64-bit Hunspell for the 64-bit b…

samdoss:
…uild

* Added the Build Platform for x64bit

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/1)


papeh:
Looks good overall. One question about output paths (see line comment). Also, do we need to add an x86 PropertyGroup for each project as well?

---

Reviewed 11 of 11 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/CSTools/lg/lg.csproj, line 71 at r1](https://reviewable.io:443/reviews/sillsdev/liblcm/1#-KmgTKLq9luNnkPsjrMk:-KmgTKLq9luNnkPsjrMl:bddv9up) ([raw file](https://github.com/sillsdev/liblcm/blob/121efc70f7fd0ac9c26075b757a5cf1545b715f6/src/CSTools/lg/lg.csproj#L71)):*
> ```XML
>   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|x64'">
>     <DebugSymbols>true</DebugSymbols>
>     <OutputPath>bin\x64\Debug\</OutputPath>
> ```

shouldn't this OutputPath be similar to the path for AnyCPU? (twice in each file)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/liblcm/1#-:-KmgT_Eh164s6gmDP1gP:b62zslr)*

ddaspit:
If we are going to update LCM to support 64-bit, then we should not add new configurations for the 64-bit platform. Instead, we should switch all of the projects to Any CPU. There is no need to have separate 32-bit and 64-bit builds. Please do not merge this PR until this is addressed. 

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/liblcm/1#-:-KmlsBsbGQzq3zaiXUJh:bjmxpmo)*

ermshiperete:
@samdoss Since your change was on your personal fork (and nobody should clone from your fork), you can simply do a hard reset to undo your change - no need to add a revert commit or merge this PR. That would be something like:

    git reset --hard origin/master

(you might have to replace `origin` with whatever your remote is called that points to `github.com/sillsdev/liblcm.git`).

This PR should be closed without merging it.

jasonleenaylor:
@ddaspit We plan to use a NuGet package for hunspell which can be either 64bit or 32bit. I suspect we will need two builds in the end.

papeh:
If we don't like the complete history, we can Squash and merge this pull request

ermshiperete:
> We plan to use a NuGet package for hunspell which can be either 64bit or 32bit. I suspect we will need two builds in the end.

We should be able to create a nuget package that contains both 64- and 32bit versions in the same package, similar to how it's done with the `Icu4c.Win.*` packages.

ddaspit:
@jasonleenaylor As long as there is no code that is 32-bit or 64-bit specific in the LCM assemblies, we should be able to set all of the assemblies to Any CPU. If it is set to Any CPU, then it will run in 32-bit in a 32-bit application and 64-bit in a 64-bit application. We just need to make sure that the correct native DLL is present when the application is installed.

jasonleenaylor:
@ddaspit Ok, so the LCM assembly would build AnyCpu and would load a dll which in the end could be either 32 or 64bit but it doesn't matter which? If we keep the name the same and make the installer choose the right version. Are we on the same page now?

ddaspit:
@jasonleenaylor Yes, I think we're on the same page now.

PR #4, opened 2017-06-30
Fix: LT-18378 Update Flex to use the 64-bit Hunspell Build

samdoss:
* Added the Nhunspell Package in LCModel Core.
* Modified in code based on the Hunspell functions.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/4)


src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
```diff
@@ -75,9 +74,10 @@ public bool Check(string word)
 		/// </summary>
 		/// <param name="word"></param>
 		/// <returns></returns>
-		private static byte[] MarshallAsUtf8Bytes(string word)
+		private static string MarshallAsUtf8Bytes(string word)
 		{
-			return Encoding.UTF8.GetBytes(Icu.Normalize(word, Icu.UNormalizationMode.UNORM_NFC) + "\0");
+			byte[] bytes = Encoding.UTF8.GetBytes(Icu.Normalize(word, Icu.UNormalizationMode.UNORM_NFC) + "\0");
```
jasonleenaylor: I'm not sure that this method is necessary anymore. The Icu.Normalize call may be all that is necessary since we don't need to pass a byte array to NHunspell.

jasonleenaylor: Ok, I now understand, the UTF8 encoding is required, and even though the NHunspell takes a string instead of bytes this is really the only way to get a UTF8 encoded string.

src/SIL.LCModel.Core/SIL.LCModel.Core.csproj
```diff
@@ -226,43 +226,43 @@
...
       <Link>icuin54.dll</Link>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
-    <Content Include="..\..\lib\downloads\x64\icutu54.dll" Condition="'$(Platform)'!='x86' and '$(OS)'=='Windows_NT'">
+    <Content Include="..\..\lib\downloads\x64\icutu54.dll" Condition="'$(Platform)'!='x86'">
       <Link>icutu54.dll</Link>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
-    <Content Include="..\..\lib\downloads\x64\icuuc54.dll" Condition="'$(Platform)'!='x86' and '$(OS)'=='Windows_NT'">
+    <Content Include="..\..\lib\downloads\x64\icuuc54.dll" Condition="'$(Platform)'!='x86'">
```
jasonleenaylor: On Linux the icu libraries are actually used from a system path. So please add back the OS condition for the icu.dlls.
After that this will be ready to merge! Good work!

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
```diff
@@ -75,9 +97,16 @@ public bool Check(string word)
...
+		#if __MonoCS__
 		private static byte[] MarshallAsUtf8Bytes(string word)
 		{
 			return Encoding.UTF8.GetBytes(Icu.Normalize(word, Icu.UNormalizationMode.UNORM_NFC) + "\0");
+		#else
+		private static string MarshallAsUtf8Bytes(string word)
+		{
+			byte[] bytes = Encoding.UTF8.GetBytes(Icu.Normalize(word, Icu.UNormalizationMode.UNORM_NFC) + "\0");
+			return Encoding.UTF8.GetString(bytes);
+		#endif
```
papeh: I think this would be clearer if the close brace `}` were inside the `#if` block

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
```diff
@@ -50,21 +63,30 @@ internal SpellEngine(string affixPath, string dictPath, string exceptionPath)
...
 				{
 					Hunspell_uninitialize(m_hunspellHandle);
 					m_hunspellHandle = IntPtr.Zero;
 					throw;
 				}
+				#else
+				if (m_hunspellHandle != null)
+				{
+					m_hunspellHandle.Dispose();
+					throw;
```
papeh: Since we're working in the area, it would be nice to change these to `throw new Exception(e);` to preserve the original stacktrace

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
```diff
@@ -107,13 +140,23 @@ public bool IsVernacular
...
 			Hunspell_free_list(m_hunspellHandle, ref pointerToAddressStringArray, resultCount);
 			return results;
+			#else
+			List<string> suggestions = m_hunspellHandle.Suggest(MarshallAsUtf8Bytes(badWord));
+			int resultCount = suggestions.Count;
+			if (resultCount == 0)
+				return new string[0];
+
+			var results = suggestions;
+			return results;
```
papeh: this entire section could be replaced with `return m_hunspellHandle.Suggest(MarshallAsUtf8Bytes(badWord))`

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
```diff
@@ -168,20 +211,35 @@ private void SetInternalStatus(string word, bool isCorrect)
 			{
 				if (IsVernacular)
 				{
+					#if __MonoCS__
```
papeh: I think this `#if` block belongs after the comment // Custom vernac....other-case matches

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
```diff
@@ -203,14 +261,19 @@ public void Dispose()
...
 		{
-			Debug.WriteLineIf(!disposing, "****** Missing Dispose() call for " + GetType().Name + ". ****** ");
+			#if __MonoCS__
 			if (m_hunspellHandle != IntPtr.Zero)
 			{
 				Hunspell_uninitialize(m_hunspellHandle);
 				m_hunspellHandle = IntPtr.Zero;
 			}
+			#else
+			Debug.WriteLineIf(!disposing, "****** Missing Dispose() call for " + GetType().Name + ". ****** ");
```
papeh: this `WriteLineIf` should always be the first line of `virtual void Dispose(bool)`

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
```diff
@@ -243,6 +306,8 @@ static string[] MarshalUnmananagedStrArray2ManagedStrArray(IntPtr pUnmanagedStri
 		private const string LibHunspell = "libhunspell";
 		private const string LibHunspellPrefix = "hunspell_";
 
+
+
```
papeh: don't need these extra lines

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
```diff
@@ -278,13 +343,18 @@ static string[] MarshalUnmananagedStrArray2ManagedStrArray(IntPtr pUnmanagedStri
 		private static int Hunspell_suggest(IntPtr handle, byte[] word, out IntPtr suggestions)
 		{
 			if (MiscUtils.IsUnix)
+			{
 				return Hunspell_suggest_unix(handle, out suggestions, word);
-			return Hunspell_suggest_win(handle, word, out suggestions);
+			}
+			else
+			{
+				return Hunspell_suggest_win(handle, word, out suggestions);
```
papeh: now that this is in an `#if __MonoCS__` block, we don't need `Hunspell_suggest_win` anymore.

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
```diff
@@ -107,13 +140,23 @@ public bool IsVernacular
...
 		{
+			#if __MonoCS__
 			IntPtr pointerToAddressStringArray;
 			int resultCount = Hunspell_suggest(m_hunspellHandle, MarshallAsUtf8Bytes(badWord), out pointerToAddressStringArray);
 			if (pointerToAddressStringArray == IntPtr.Zero)
 				return new string[0];
 			var results = MarshalUnmananagedStrArray2ManagedStrArray(pointerToAddressStringArray, resultCount);
 			Hunspell_free_list(m_hunspellHandle, ref pointerToAddressStringArray, resultCount);
 			return results;
+			#else
```
papeh: To clarify my comment below, the `#if __MonoCS__` half of this block can stay as-is, but the `#else` portion can be simplified to one line.

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
samdoss: Applied (throw new Exception(e);) the change.

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
samdoss: Modified the code based on the suggestion.

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
samdoss: Changed in the code.

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
samdoss: Updated based on the comment.

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
samdoss: Moved the Debug.WriteLineIf.

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
samdoss: removed the line spacing.

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
samdoss: Removed this method Hunspell_suggest_win.

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
samdoss: Modified all the code based on the comments and committed the changes in GITHUB.

PR #5, opened 2017-07-14
Various small improvements

ermshiperete:
- Re-throw exception instead of creating a new one in SpellEngine
- Replace two obsolete method calls
- Update tools to use .NET 4.5 instead of .NET 2.0

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/5)

### DIFF
--- a/src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
+++ b/src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
@@ -72,7 +72,7 @@ namespace SIL.LCModel.Core.SpellChecking
 #else
 				m_hunspellHandle?.Dispose();
 #endif
-				throw new Exception(e.Message);
+				throw;
 			}
 		}
 
--- a/src/SIL.LCModel/DomainImpl/OverridesCellar.cs
+++ b/src/SIL.LCModel/DomainImpl/OverridesCellar.cs
@@ -2665,7 +2665,7 @@ namespace SIL.LCModel.DomainImpl
 				// The XML is from a file shipped with FieldWorks. It is quite likely multiple users
 				// of a project could independently add the same items, so we create them with fixed guids
 				// so merge will recognize them as the same objects.
-				string guid = XmlUtils.GetAttributeValue(item.SelectSingleNode("fs"), "typeguid");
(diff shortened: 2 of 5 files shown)


src/SIL.LCModel/DomainImpl/OverridesCellar.cs
```diff
@@ -2665,7 +2665,7 @@ public IFsFeatStrucType GetOrCreateFeatureTypeFromXml(XmlNode item)
 				// The XML is from a file shipped with FieldWorks. It is quite likely multiple users
 				// of a project could independently add the same items, so we create them with fixed guids
 				// so merge will recognize them as the same objects.
-				string guid = XmlUtils.GetAttributeValue(item.SelectSingleNode("fs"), "typeguid");
```
jasonleenaylor: GetAttributeValue would throw if there was no attribute present. Given there is no null check on the guid here this should really be a GetRequiredAttributeValue if such a method exists.

src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
```diff
@@ -72,7 +72,7 @@ internal SpellEngine(string affixPath, string dictPath, string exceptionPath)
 #else
 				m_hunspellHandle?.Dispose();
 #endif
-				throw new Exception(e.Message);
+				throw;
```
jasonleenaylor: So this is the same as throw e? I guess I didn't realize that.

ermshiperete: I just refreshed my memory - to re-throw the exception just `throw` is the preferred way because it preserves the original call stack (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/throw#re-throwing-an-exception)

src/SIL.LCModel/DomainImpl/OverridesCellar.cs
ermshiperete: Done. Thanks for catching this!

jasonleenaylor:
:lgtm:

---

Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/liblcm/5#-:-KpzTfdpG5t8dq1Kc3nm:bnfp4nl)*

PR #7, opened 2017-07-26
Fix LT-18451:Homograph number does not appear with entries

vkarthim:
*Added code to include Homograph number in Grammer-Category
*Added TODO comment which should do GrammerSketch work

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/7)


jasonleenaylor:
This work was done on another branch and is a first commit toward enabling homograph numbers in some locations where users wanted it.

PR #21, opened 2017-11-21
Choose available libhunspell version at runtime

marksvc:
* To support Ubuntu 18.04 (libhunspell 1.6) at the same time as 16.04
and 14.04 (libhunspell 1.3).
* Also falls back on libhunspell.so, provided by package libhunspell-
dev.
* Fixes crash LT-18708.
* There are many ways to implement this. If our use of libhunspell
grows in complexity, another way may be more suitable.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/21)

### DIFF
--- a/src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
+++ b/src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
@@ -25,22 +25,23 @@ namespace SIL.LCModel.Core.SpellChecking
 		/// </summary>
 		internal string ExceptionPath { get; set; }
 		#if __MonoCS__
-			private IntPtr m_hunspellHandle;
+		private IntPtr m_hunspellHandle;
+		private NativeLibhunspell _libhunspell = new NativeLibhunspell();
 		#else
-			private Hunspell m_hunspellHandle;
+		private Hunspell m_hunspellHandle;
 		#endif
-		
+
 		internal SpellEngine(string affixPath, string dictPath, string exceptionPath)
 		{
 			try
 			{
 				#if __MonoCS__
-				m_hunspellHandle = Hunspell_initialize(MarshallAsUtf8Bytes(affixPath), MarshallAsUtf8Bytes(dictPath));
+				m_hunspellHandle = _libhunspell.Hunspell_initialize(MarshallAsUtf8Bytes(affixPath), MarshallAsUtf8Bytes(dictPath));
 				#else
 				m_hunspellHandle = new Hunspell(affixPath, dictPath);
 				#endif
 				ExceptionPath = exceptionPath;
-				
+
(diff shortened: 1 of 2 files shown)


src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs
```diff
@@ -291,47 +292,442 @@ static string[] MarshalUnmananagedStrArray2ManagedStrArray(IntPtr pUnmanagedStri
...
+		[DllImport("libdl.so", SetLastError = true)]
+		private static extern int dlclose(IntPtr handle);
+
+		private ILibHunspell nativeLibrary;
+
+		private ILibHunspell Library
+		{
+			get
+			{
+				if (nativeLibrary == null)
```
jasonleenaylor: Please refactor this to reduce the nesting, otherwise it all looks good.

ermshiperete:
:lgtm: Looks good, although it has a bit too much duplication for my taste, but at least it solves the immediate problem. It still can lead to a crash I think if the Remove method isn't included in the .so, but maybe we don't call that method on Linux.

In the long run it would be nice to get rid of the `#if`s - there's no `__MonoCS__` with Mono 5, and it would simplify creating a nuget package if we would have the same binary on Linux and Windows. But that can be a later change if we want to.

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/liblcm/21#-:-KzXeHabBtzyP7HWDCcq:b-n4xhs5)*

jasonleenaylor:
*[src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs, line 317 at r1](https://reviewable.io:443/reviews/sillsdev/liblcm/21#-KzVYhTLVtebLdDzhC37-r1-317:-KzyTzPnrxjNDBQJZWb4:b-bjpj7r) ([raw file](https://github.com/sillsdev/liblcm/blob/a210e7a0a6b5840fd4bd8eb2b0f59d342665f9e8/src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs#L317)):*

Please refactor this to reduce the nesting, otherwise it all looks good.
</blockquote></details>

Better.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/liblcm/21)*

jasonleenaylor:
:lgtm:

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/liblcm/21#-:-KzyUH3S1WA_iIySiWDE:bnfp4nl)*

PR #26, opened 2017-12-01
 Fix LT-18253:Variant Type not consistently set to 'Unspecified Variant'

vkarthim:
*Changed the logic to set 'Unspec Variant' as default
*Find default type using Guid
*Assert added to the existing test case for this issue.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/26)


src/SIL.LCModel/DomainImpl/OverridesLing_Lex.cs
```diff
@@ -1324,7 +1324,9 @@ public ILexEntryRef MakeVariantOf(IVariantComponentLexeme componentLexeme, ILexE
 			}
 			else
 			{
-				ler.VariantEntryTypesRS.Add(Cache.LangProject.LexDbOA.VariantEntryTypesOA.PossibilitiesOS[0] as ILexEntryType);
+				int defVariantIdx = Cache.LangProject.LexDbOA.VariantEntryTypesOA.PossibilitiesOS.Where(e =>
```
jasonleenaylor: I was thinking that this could be simplified, but the way it is written has the effect of avoiding any extra conditionals for unit tests where the unspecified variant type might not be in the list.

PR #27, opened 2017-12-07
Fix LT-18781:Crash appears on "Lexicon Edit" pane when changing category

vkarthim:
*Added one more condition to set null
*InflFeatsOA also be null when invalid features.
*Added test case OnRefresh_SetNullInflFeatsOAWhenInvalidFeature

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/27)

### DIFF
--- a/tests/SIL.LCModel.Tests/DomainImpl/LingTests.cs
+++ b/tests/SIL.LCModel.Tests/DomainImpl/LingTests.cs
@@ -353,6 +353,7 @@ namespace SIL.LCModel.DomainImpl
 			Assert.That(mdsm.FeaturesTSS.Text, Is.EqualTo(featSpec6.ShortNameTSS.Text),
 				"4 with one features (MsInflFeatsOA). FeaturesTSS should not be an empty string");		}
 
+
 		/// <summary>
 		/// Test the ExceptionFeaturesTSS virtual property.
 		/// </summary>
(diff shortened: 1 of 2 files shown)


tests/SIL.LCModel.Tests/DomainImpl/LingTests.cs
```diff
@@ -2064,6 +2065,71 @@ public void MoStemMsa_EqualMsa()
...
+			doc.LoadXml(MoreCellarTests.ksFS1);
+			XmlNode itemNeut = doc.SelectSingleNode("/item/item[3]");
+			XmlNode itemFem = doc.SelectSingleNode("/item/item[2]");
+			var featStruc1 = Cache.ServiceLocator.GetInstance<IFsFeatStrucFactory>().Create();
+			var featStruc2 = Cache.ServiceLocator.GetInstance<IFsFeatStrucFactory>().Create();
+			infAffixMsa1.InflFeatsOA = featStruc1;
+			infAffixMsa2.InflFeatsOA = featStruc2;
+			infAffixMsa1.InflFeatsOA.AddFeatureFromXml(itemNeut, Cache.LangProject.MsFeatureSystemOA);
+			infAffixMsa2.InflFeatsOA.AddFeatureFromXml(itemFem, Cache.LangProject.MsFeatureSystemOA);
+
```
jasonleenaylor: Add an assert here to verify that infAffixMsa1.InflFeatsOA is not null before we change the POS.

tests/SIL.LCModel.Tests/DomainImpl/LingTests.cs
```diff
@@ -2064,6 +2065,71 @@ public void MoStemMsa_EqualMsa()
...
+			// reset POS
```
jasonleenaylor: Change comment to // change to POS with incompatible features

tests/SIL.LCModel.Tests/DomainImpl/LingTests.cs
vkarthim: Ok

tests/SIL.LCModel.Tests/DomainImpl/LingTests.cs
vkarthim: Ok

PR #31, opened 2018-01-18
Allow using a Tss as a Morph in ParagraphParserTests

papeh:
Also, more sensible method order in WritingSystemServices

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/31)

### DIFF
--- a/src/SIL.LCModel/DomainServices/WritingSystemServices.cs
+++ b/src/SIL.LCModel/DomainServices/WritingSystemServices.cs
@@ -577,11 +577,6 @@ namespace SIL.LCModel.DomainServices
 
 		}
 
-		/*----------------------------------------------------------------------------------------------
-
-			@param ws
-			@return a real writing system
-		----------------------------------------------------------------------------------------------*/
 		/// <summary>
 		/// Convert a writing system (could be magic) to a real writing system.
 		/// </summary>
@@ -602,11 +597,6 @@ namespace SIL.LCModel.DomainServices
 			return retWs;
 		}
 
-		/*----------------------------------------------------------------------------------------------
-
-			@param ws
-			@return a real writing system
-		----------------------------------------------------------------------------------------------*/
 		/// <summary>
 		/// Convert a writing system (could be magic) to a real writing system.
 		/// </summary>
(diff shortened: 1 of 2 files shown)


src/SIL.LCModel/DomainServices/WritingSystemServices.cs
```diff
@@ -1101,16 +1076,10 @@ public static bool TryWs(LcmCache cache, int wsPreferred, int wsSecondary, int h
...
-		/// <param name="ws">The ws.</param>
-		/// <param name="hvoOwner">The hvo owner.</param>
-		/// <param name="flidOwning">The flid owning.</param>
-		/// <param name="actualWs">The actual ws.</param>
-		/// <param name="tssResult">The TSS result.</param>
 		/// <returns>
-		/// true if ws is nonzero and tssResult has content (length greater than 0)
+		/// true if ws is real and tssResult has content (length greater than 0)
 		/// </returns>
 		public static bool TryWs(LcmCache cache, int ws, int hvoOwner, int flidOwning, out int actualWs, out ITsString tssResult)
```
papeh: I briefly considered moving these up with the other helper method, and would still be glad to if prompted, but figured that would be noisier than some would prefer.

tests/SIL.LCModel.Tests/DomainServices/ParagraphParserTests.cs
```diff
@@ -687,6 +687,11 @@ public virtual IWfiAnalysis BreakIntoMorphs(int iSegment, int iSegForm, ArrayLis
 						// just set the Form.
 						wmb.Form.SetVernacularDefaultWritingSystem((string)morphForm);
 					}
+					else if (morphForm is ITsString)
+					{
+						var tssMf = (ITsString)morphForm;
+						wmb.Form.set_String(TsStringUtils.GetWsAtOffset(tssMf, 0), tssMf);
+					}
```
papeh: https://gerrit.lsdev.sil.org/#/c/5799/10/Src/LexText/Interlinear/ITextDllTests/InterlinearExporterTests.cs@228 requires this.

PR #39, opened 2018-02-15
LT-18917: Data Migration 7000072 (Reverse Reversals)

papeh:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/39)


papeh:
Before this commit, I got a yellow box opening an old FLEx project with reversals. Now, I get no yellow box, and Reversals seem to be properly connected, but I see only headwords in the XHTML (this could be a configuration problem, but resetting to factory defaults doesn't solve it, and I haven't looked at the migration log)

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/liblcm/39#-:-L5Ql5fuqXhFu9J2qYg1:byigok6)*

PR #56, opened 2018-05-23
Validate pronunciation WS during initialization

RaymondLuong3:
* Removes WS from our list if it is not recognized by our project.
* Partially fixes LT-19152

### DIFF
--- a/src/SIL.LCModel/DomainImpl/OverridesLangProj.cs
+++ b/src/SIL.LCModel/DomainImpl/OverridesLangProj.cs
@@ -1,4 +1,4 @@
-// Copyright (c) 2002-2017 SIL International
+// Copyright (c) 2002-2018 SIL International
 // This software is licensed under the LGPL, version 2.1 or later
 // (http://www.gnu.org/licenses/lgpl-2.1.html)
 //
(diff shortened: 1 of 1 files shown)


RaymondLuong3:
That is a good point. I will see if I can report it in a message box.

src/SIL.LCModel/DomainImpl/OverridesLangProj.cs
```diff
@@ -907,6 +907,31 @@ private void InitializePronunciationWritingSystems()
...
+		/// <returns>True if a valid writing system is found, otherwise false.</returns>
+		private bool CheckValidPronunciationWritingSystem()
+		{
+			CoreWritingSystemDefinition ws = null;
+			while (m_currentPronunciationWritingSystems.Count > 0)
+			{
+				try
+				{
+					ws = m_currentPronunciationWritingSystems[0];
+					return true;
```
jasonleenaylor: So what if there is one good ws, followed by one bad ws?

RaymondLuong3: Good catch! I am finding that The dictionary uses the list of CurrentPronunciationWritingSystems, which accesses this member. And if the ws is invalid, I get another crash. I will make the adjustment.

PR #62, opened 2018-07-18
Fix buffer overrun (reading) during marshalling

jasonleenaylor:
* const OLECHAR * params require null terminators
  switch to const BSTR to avoid buffer overrun

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/62)

### DIFF
--- a/src/SIL.LCModel.Core/KernelInterfaces/TextServ.idh
+++ b/src/SIL.LCModel.Core/KernelInterfaces/TextServ.idh
@@ -987,7 +987,7 @@ DeclareInterface(TsPropsFactory, Unknown, FF3D947F-1D35-487B-A769-5B6C68722054)
 ----------------------------------------------------------------------------------------------*/
 // REVIEW: ShonK, JeffG(JeffG): Should TsStrBldr derive off of TsStrBase?  and
 // Should Strbuilder(s) accept TsPropsBldrs
-DeclareInterface(TsStrBldr, Unknown, 71D84C4A-10A2-4A7C-B58C-D01386DAE5F7)
+DeclareInterface(TsStrBldr, Unknown, 35C5278D-2A52-4B54-AB13-B6E346B301BA)
 {
 	// With the exception of the Replace*, PushEcoding, GetString methods, all other methods are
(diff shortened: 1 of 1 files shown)


ermshiperete:
Just had another thought on my way home. 

Liblcm has a C# implementation of `ITsStrBldr`. I'm not sure if that's exposed to COM. In that case we would definitely have a marshalling problem when `ReplaceRgch` gets called from C++ code and the strings are not null terminated. 

Then I'd suggest to get rid of `ReplaceRgch` altogether, and also remove the C++ implementation of TsStrBldr (and related classes that are now also implemented in C#). I'd think/hope that the C++ would/should use the C# implementation.

---

Review status: all files reviewed, 2 unresolved discussions (waiting on @jasonleenaylor)

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/62#-:-LHidiWs5J8qBizC5Dyt:b-slx75s)*

jasonleenaylor:
___
*[src/SIL.LCModel.Core/KernelInterfaces/TextServ.idh, line 1066 at r1](https://reviewable.io/reviews/sillsdev/liblcm/62#-LHiNGGVB3ad_cmxGdWb:-LHn9h-u0JOR6s2u4C-P:b-9vk0ry) ([raw file](https://github.com/sillsdev/liblcm/blob/81bbff4676827feeee3d9507ad0495fc4961f4ad/src/SIL.LCModel.Core/KernelInterfaces/TextServ.idh#L1066)):*
<details><summary><i>Previously, ermshiperete (Eberhard Beilharz) wrote…</i></summary><blockquote>

This change makes this method basically the same as `Replace()`, so the only question I have here is if we should remove `ReplaceRgch` from the COM interface.

I think I'd prefer if we'd get rid of `ReplaceRgch` since we have to change every place that calls it anyways, or if we want to keep this method, leave it as it was previously and in the C++ code replace the use of `ReplaceRgch` with `Replace`.
</blockquote></details>

The removing of ReplaceRgch is proving a painful amount of work in the C++. I'm reconsidering my options. Perhaps I should look into modifying the code generation. Now to find where that code is...

PR #64, opened 2018-07-31
Fix: LT-19199 Crash while creating a new lexical entry

JamesPrabu:
* Added the code to remove reference

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/64)

### DIFF
--- a/src/SIL.LCModel/DomainImpl/OverridesLing_Lex.cs
+++ b/src/SIL.LCModel/DomainImpl/OverridesLing_Lex.cs
@@ -3298,6 +3298,17 @@ namespace SIL.LCModel.DomainImpl
 		}
 
 		/// <summary>
+		/// Remove one reference to the target object from one of your atomic reference properties.
+		/// </summary>
+		internal override void RemoveAReferenceCore(ICmObject target)
+		{
+			var flid = this.Cache.CustomProperties.Where(x => x.Value == target && x.Key.Item1 == this).Select(y => y.Key.Item2).FirstOrDefault();
+			if (flid > 0)
+				SetNonModelPropertyForSDA(flid, null, true);
+			base.RemoveAReferenceCore(target);
+		}
+
+		/// <summary>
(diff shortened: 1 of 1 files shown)


src/SIL.LCModel/DomainImpl/CmObject.cs
```diff
@@ -2894,7 +2894,7 @@ internal static GenDate GetGenDateFromInt(int nVal)
 		/// <param name="flid">Custom field id.</param>
 		/// <param name="newValue">New value. (May be null.)</param>
 		/// <param name="useAccessor">'true' to send to UOW. 'false' to not send to the UOW.</param>
-		private void SetCustomPropertyForSDA(int flid, object newValue, bool useAccessor)
+		public void SetCustomPropertyForSDA(int flid, object newValue, bool useAccessor)
```
jasonleenaylor: I didn't realize that you needed to make this public. There is probably a better way of getting this change made through a method that is already internal or protected.

src/SIL.LCModel/DomainImpl/OverridesLing_Lex.cs
```diff
@@ -3297,6 +3297,16 @@ public string ReferenceName
...
 
+		/// <summary>
+		/// Remove one reference to the target object from one of your atomic reference properties.
+		/// </summary>
+		internal override void RemoveAReferenceCore(ICmObject target)
+		{
+			var flid = this.Cache.CustomProperties.Where(x => x.Value == target && x.Key.Item1 == this).Select(y => y.Key.Item2).FirstOrDefault();
+			if (flid > 0)
+				SetNonModelPropertyForSDA(flid, null, true);
+		}
```
jasonleenaylor: James, I think you should add one line to the end of the method:
This shouldn't change any behavior, but it makes it consistent and might avoid a future bug.
base.RemoveAReferenceCore(target);

JamesPrabu: Jason, I added base.RemoveAReferenceCore(target).

PR #69, opened 2018-08-28
 Allow to create picture with existing GUID

ermshiperete:
The CmPictureFactory so far lacks the ability to create a picture from an existing GUID. However this is needed for LanguageForge S/R, so this change adds this.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/69)


jasonleenaylor:
:LGTM:

PR #80, opened 2018-11-12
Make use of custom icu.net optional

ermshiperete:
Also move methods that access ICU to icu.net


jasonleenaylor:
___
*[lib/teamcity.dep, line 1 at r1](https://reviewable.io/reviews/sillsdev/liblcm/80#-LRIBCur1yzpLgdGy9rr:-LRSVakHCrIXZNFE4Izz:bj00zka) ([raw file](https://github.com/sillsdev/liblcm/blob/eaf602dd249099806444d1b11a81083fc10ba52c/lib/teamcity.dep#L1)):*
<details><summary><i>Previously, ermshiperete (Eberhard Beilharz) wrote…</i></summary><blockquote>

For the record: The current setup will produce warnings like `warning MSB3270: There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture of the reference "SIL.LCModel.Utils.Tests, Version=9.0.0.0, Culture=neutral, processorArchitecture=MSIL", "x86". This mismatch may cause runtime failures. Please consider changing the targeted processor architecture of your project through the Configuration Manager so as to align the processor architectures between your project and references, or take a dependency on references with a processor architecture that matches the targeted processor architecture of your project.` when used in FieldWorks and elsewhere.
</blockquote></details>

I understand, I didn't like it either. The problem is that we have to make the change all the way down the line. Every single use of an unmanaged library in every single managed library that we build or use must be changed to bring and load both versions. This was too much for our team to take on at the time. And likely too much for you to take on now.

PR #89, opened 2019-01-11
Fix LT-19219: Crash when looking at text with surrogate pairs

JamesPrabu:
* Added code to prevent the crash

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/89)

### DIFF
--- a/src/SIL.LCModel.Core/Text/TextExtensions.cs
+++ b/src/SIL.LCModel.Core/Text/TextExtensions.cs
@@ -355,6 +355,8 @@ namespace SIL.LCModel.Core.Text
 			if (ich < 0 || ich >= tss.Length)
 				throw new ArgumentOutOfRangeException("ich");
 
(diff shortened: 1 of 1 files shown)


src/SIL.LCModel.Core/Text/TextExtensions.cs
```diff
@@ -355,6 +355,7 @@ public static bool IsCharWordForming(this ITsString tss, int ich, ILgWritingSyst
 			if (ich < 0 || ich >= tss.Length)
 				throw new ArgumentOutOfRangeException("ich");
 
+			if (char.IsLowSurrogate(tss.Text[ich])) --ich;
```
jasonleenaylor: please move the --ich to the next line.

If there is already a unit test that covers this function we should add a new test to document this fix.

JamesPrabu: We moved --ich to next line.

PR #92, opened 2019-02-14
Fix performance bug when pasting large amounts of text

jasonleenaylor:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/92)

### DIFF
--- a/src/SIL.LCModel.Core/Text/TsStrBldr.cs
+++ b/src/SIL.LCModel.Core/Text/TsStrBldr.cs
@@ -1,9 +1,10 @@
-﻿// Copyright (c) 2016 SIL International
+// Copyright (c) 2016 SIL International
 // This software is licensed under the LGPL, version 2.1 or later
 // (http://www.gnu.org/licenses/lgpl-2.1.html)
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using SIL.Extensions;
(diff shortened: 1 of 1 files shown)


src/SIL.LCModel.Core/Text/TsStrBldr.cs
```diff
@@ -319,5 +319,59 @@ public void Clear()
...
+			}
+
+			public void Clear()
+			{
+				_cachedString = null;
+				_stringBuilder.Clear();
+			}
+
+			public void Remove(int ichMin, int i)
+			{
```
ermshiperete: You could add an assert like `Debug.Assert(_cachedString == null || _stringBuilder.Length == 0);`

src/SIL.LCModel.Core/Text/TsStrBldr.cs
```diff
@@ -319,5 +319,59 @@ public void Clear()
...
+				{
+					_stringBuilder.Append(_cachedString);
+					_cachedString = null;
+				}
+
+				_stringBuilder.Remove(ichMin, i);
+			}
+
+			public void Insert(int ichMin, string insertText)
+			{
```
ermshiperete: You could add an assert like `Debug.Assert(_cachedString == null || _stringBuilder.Length == 0);`

jasonleenaylor:
Yes, when we are pasting a large string into the baseline it ends up calling .ToString() on the builder multiple times for every character which is in the pasted text. For 'small' strings this is fine, but for large strings it is expensive to make the copies.

jasonleenaylor:
LGTM based on Eberhard's coment

PR #97, opened 2019-03-29
Generate expected label for writing systems

jasonleenaylor:
* Fix https://jira.sil.org/browse/LT-19527
* Add unit test for display label expectations

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/97)


jasonleenaylor:
___
*[src/SIL.LCModel/DomainServices/DataMigration/DataMigration7000071.cs, line 160 at r2](https://reviewable.io/reviews/sillsdev/liblcm/97#-Ld5vkYW-Y6UMtncAeP-:-Ld9wMHA4w8LUD10N1Qv:b-tv6gzc) ([raw file](https://github.com/sillsdev/liblcm/blob/d67cfe5ce2ee2746019d65d0738f4f88faf4aaec/src/SIL.LCModel/DomainServices/DataMigration/DataMigration7000071.cs#L160)):*

why are we dropping the canonicalization?
</blockquote></details>

On purpose

PR #98, opened 2019-05-06
[API] Filter non-ASCII characters out of project names

papeh:
* Add an optional parameter to StringUtils.FilterForFileName to remove
  non-ASCII characters
* In spite of our best efforts, these characters still cause occasional
  S/R problems in mercurial, so we will avoid using them
* Addresses https://jira.sil.org/browse/LT-19693

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/98)


papeh:
Another idea was to add the Project Name as an optional parameter to MiscUtils.GetInvalidProjectNameChars and include the Project Name's non-ASCII characters in the result

PR #100, opened 2019-05-17
LT-19685: Prevent French from appearing magically

papeh:
- Remove remaining references to French in the new project template
- This should prevent French from magically appearing in users' projects when they have not added it. This also prevents French from being added to the Global WS Repo unless a user adds it to a project.
- https://jira.sil.org/browse/LT-19685

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/100)

### DIFF
--- a/src/SIL.LCModel/Templates/NewLangProj.fwdata
+++ b/src/SIL.LCModel/Templates/NewLangProj.fwdata
@@ -1711,9 +1711,6 @@
 <GenreList>
 <objsur guid="c06b099c-ea5e-11de-890c-0013722f8dec" t="o" />
 </GenreList>
-<HomographWs>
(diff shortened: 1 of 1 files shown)


src/SIL.LCModel/Templates/NewLangProj.fwdata
```diff
@@ -1712,7 +1712,7 @@
 <objsur guid="c06b099c-ea5e-11de-890c-0013722f8dec" t="o" />
 </GenreList>
 <HomographWs>
-<Uni>fr</Uni>
+<Uni>en</Uni>
```
jasonleenaylor: If we can just leave this out it would be better.

PR #103, opened 2019-06-27
LT-19632: Share WS Data w/ SLDR by default

papeh:
If the user has not set the FEEDBACK=off environment variable,
then enable sharing of Writing System data with SLDR for new projects.
https://jira.sil.org/browse/LT-19632

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/103)

### DIFF
--- a/src/SIL.LCModel/LcmCache.cs
+++ b/src/SIL.LCModel/LcmCache.cs
@@ -417,6 +417,8 @@ namespace SIL.LCModel
 					progressDlg.Step(0);
 				}
 
+				SetDefaultProjectSettings(cache.ServiceLocator.DataSetup.ProjectSettingsStore);
+
 				cache.ActionHandlerAccessor.BeginNonUndoableTask();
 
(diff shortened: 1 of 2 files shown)


jasonleenaylor:
___
*[src/SIL.LCModel/LcmCache.cs, line 534 at r1](https://reviewable.io/reviews/sillsdev/liblcm/103#-LiOYXgN0LaChRtvETlf:-LiOYXgO-SePY-okly_d:bas0fru) ([raw file](https://github.com/sillsdev/liblcm/blob/0eca3414dfcd1710fe4944b2f4e3419bc2fb164a/src/SIL.LCModel/LcmCache.cs#L534)):*
> ```Smalltalk
> 		private static void SetDefaultProjectSettings(LcmCache cache)
> 		{
> 			var feedbackEnvVar = Environment.GetEnvironmentVariable("FEEDBACK");
> ```

Let's skip this, if/when we start pushing to SLDR we can honor the FEEDBACK  setting in that code.

PR #110, opened 2019-12-04
Bugfix/fix assembly redirect

jasonleenaylor:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/110)


jasonleenaylor:
You are correct. However this is a very confusing situation, which has bit me more than once. I thought it had changed since the information visible when inspecting the .dll on windows doesn't expose this version.

jasonleenaylor:
When we get errors loading .icu this is the first thing that jumps out as a possible problem. The only way I can see what the actual assembly version is as a consumer is to open the file in dotpeek.

PR #128, opened 2020-03-25
Fix crash from casting TsTextProps (LT-20002)

JohnThomson:
Seems when a C# object has been marshaled for COM  and passed to C++ and back, a simple cast will not work to get back to the implementing C# object. I can't find what will, so took another approach.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/128)


JohnThomson:
@ddaspit Jason says you figured out the COM marshaling, so you may have a better solution.

PR #155, opened 2020-09-30
Fix crash shutting down

ermshiperete:
This change fixes a crash that we got in FW after we got another exception and displayed a green screen and the user clicked ok. In that case `ShutdownInteral()` got called before we called `Commit()`.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/155)


ermshiperete:
No, doesn't really help. Just moves the crash to a different place...

PR #179, opened 2021-02-04
LT-20408: Allow adding valid characters with diacritics

papeh:
* Upgrade to the latest version of ICU.NET, which has the
proper return type for Normalizer2.GetCombiningClass

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/179)


papeh:
will be included in the next merge.

PR #180, opened 2021-02-04
LT-20408: Allow adding valid characters with diacritics

papeh:
* Cast the return type of Normalizer2.GetCombiningClass to byte.
This will have to do until we can upgrade to the latest version
of ICU.NET, which has the proper return type.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/180)


papeh:
not needed; ICU is fixed

PR #183, opened 2021-03-04
[API] LT-20432: Option to Restore Hidden Writing Systems

papeh:
* Add a method to search for all WS's in a project
* Make ITsStrBldr.Append methods return `this` to allow call chaining

+semver:minor

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/183)

### DIFF
--- a/src/SIL.LCModel/DomainServices/WritingSystemServices.cs
+++ b/src/SIL.LCModel/DomainServices/WritingSystemServices.cs
@@ -1529,38 +1529,20 @@ namespace SIL.LCModel.DomainServices
 		/// </returns>
 		public static ISet<int> FindAllWritingSystemsWithText(LcmCache cache)
 		{
-			return FindAllWritingSystemsWithText(cache, out _, out _, out _);
-		}
-
-		/// <param name="cache"/>
-		/// <param name="monoStringWSs">the handles of writing systems that have text in monolingual strings (such as Baseline Texts)</param>
-		/// <param name="multiStringWSs">the handles of writing systems that have MultiString alternatives (such as headwords and glosses</param>
-		/// <param name="embeddedWSs">the handles of writing systems that have text embedded in MultiStrings (of their own or another WS)</param>
-		/// <returns>
-		/// the handles of all Writing Systems that have text in the project, including text embedded in other writing systems' strings
-		/// </returns>
-		public static ISet<int> FindAllWritingSystemsWithText(LcmCache cache,
-			out ISet<int> monoStringWSs, out ISet<int> multiStringWSs, out ISet<int> embeddedWSs)
-		{
-			var strHandles = new HashSet<int>();
-			var multiStrHandles = new HashSet<int>();
-			var inMultiStrHandles = new HashSet<int>();
-			StringServices.CrawlStrings(cache, str => FindAllWritingSystemsInTsString(str, strHandles), multiStr =>
+			var allHandles = new HashSet<int>();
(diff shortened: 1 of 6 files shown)


src/SIL.LCModel/DomainServices/WritingSystemServices.cs
```diff
@@ -1524,14 +1525,54 @@ internal static void UpdateWritingSystemListField(LcmCache cache, ICmObject obj,
...
+		/// <param name="cache"/>
+		/// <param name="wholeStringWSs">the handles of writing systems that have whole strings (that could have slices)</param>
+		/// <param name="embeddedWSs">the handles of writing systems that have data embedded in other writing systems' strings</param>
+		/// <returns>the handles of all Writing Systems that have data in the project</returns>
+		public static ISet<int> FindAllWritingSystemsWithData(LcmCache cache, out ISet<int> wholeStringWSs, out ISet<int> embeddedWSs)
+		{
+			var strHandles = new HashSet<int>();
+			var multiStrHandles = new HashSet<int>();
+			var inMultiStrHandles = new HashSet<int>();
+			StringServices.CrawlStrings(cache, str => FindAllWritingSystemsInTsString(str, strHandles), multiStr =>
```
papeh, line 1537: non-multi-strings could be part of:
* cell (column filter)—who cares?
* StTxtPara—has PreferredWsId ("en")
* PunctuationForm (has PreferredWsID; do we care?)

papeh: Perhaps we should return only a single list, especially for MVP.

papeh: For monolingual strings, determining the "outer" WS is more trouble than it's presently worth (and may be impossible). The other helpful distinction would be vernacular vs analysis, but that'd also be difficult to determine in our countless cases. We need this to offer a feature as a convenience to users who may have hidden WS's and now want to re-show them. Users ought to be able to guess whether a WS should be vern or analysis (and if data is in the wrong place, that's a separate problem).

jasonleenaylor (approved):
This looks very thorough. :LGTM:

> Reviewed 1 of 5 files at r1, 1 of 4 files at r2, 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/183#-:-MVHwdZnAS7AqCSgWrz8:bj01qd9)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @papeh)

___
*[src/SIL.LCModel/DomainServices/WritingSystemServices.cs, line 1537 at r1](https://reviewable.io/reviews/sillsdev/liblcm/183#-MUy9yaRRDY6v1tTQWIG-r1-1537:-MVHwkQS3bqHlKeofwOY:b-bj090f) ([raw file](https://github.com/sillsdev/liblcm/blob/df6b7760db5cf5d9bfde183965528b9eafb719ae/src/SIL.LCModel/DomainServices/WritingSystemServices.cs#L1537)):*

For monolingual strings, determining the "outer" WS is more trouble than it's presently worth (and may be impossible). The other helpful distinction would be vernacular vs analysis, but that'd also be difficult to determine in our countless cases. We need this to offer a feature as a convenience to users who may have hidden WS's and now want to re-show them. Users ought to be able to guess whether a WS should be vern or analysis (and if data is in the wrong place, that's a separate problem).
</blockquote></details>

I agree with the single list for MVP.

PR #194, opened 2021-09-03
[API] Add overload of MockFileOS.AddFile with default Encoding.UTF8

papeh:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/194)

### DIFF
--- a/tests/SIL.LCModel.Utils.Tests/MockFileOs.cs
+++ b/tests/SIL.LCModel.Utils.Tests/MockFileOs.cs
@@ -1,4 +1,4 @@
-// Copyright (c) 2009-2017 SIL International
+// Copyright (c) 2009-2021 SIL International
 // This software is licensed under the LGPL, version 2.1 or later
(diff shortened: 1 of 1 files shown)


ermshiperete:
I don't see much benefit in this new overload - wouldn't it be just as easy to explicitly pass UTF8 as parameter value? And where is it called/used?

papeh:
Other methods in this class have similar overloads. Not having to pass the argument each time makes client code cleaner.

This would be immediately used by FwUpdaterTests
https://github.com/sillsdev/FieldWorks/blob/bf1e41d7e4b71df4d58ee23c2382a59f471970d1/Src/Common/FwUtils/FwUtilsTests/FwUpdaterTests.cs#L515

jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/194)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @papeh)

PR #221, opened 2022-02-23
Migration to .NET Standard

josephmyers:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/221)


papeh (commented):
seems reasonable so far.

> Reviewed 1 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/221#-:-MwbSdmB1mVKxUAvBu6D:b-eapa0r)* status: 1 of 15 files reviewed, 1 unresolved discussion (waiting on @josephmyers and @papeh)

___
*[src/SIL.LCModel/SIL.LCModel.csproj, line 87 at r1](https://reviewable.io/reviews/sillsdev/liblcm/221#-MwbSHk8CZyqpceDTQyP:-MwbSHk8CZyqpceDTQyQ:b-i7rudy) ([raw file](https://github.com/sillsdev/liblcm/blob/9abb84b603635746520576d9ac710bb13a78d949/src/SIL.LCModel/SIL.LCModel.csproj#L87)):*
> ```XML
> 
>   <PropertyGroup>
>     <OutDir Condition="'$(OutDir)' == ''">../../artifacts/$(Configuration)/$(TargetFramework)/</OutDir>
> ```

Nice. This will make things more maintainable. 😎



_Code quote:_
```XML
$(TargetFramework)
```

papeh (commented):
> Reviewed 1 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/221)* status: 1 of 16 files reviewed, 1 unresolved discussion (waiting on @josephmyers and @papeh)

___
*[src/SIL.LCModel/SIL.LCModel.csproj, line 87 at r1](https://reviewable.io/reviews/sillsdev/liblcm/221#-MwbSHk8CZyqpceDTQyP:-Mx0AWRg7x7QvpzI1Wca:b-nw2wrn) ([raw file](https://github.com/sillsdev/liblcm/blob/9abb84b603635746520576d9ac710bb13a78d949/src/SIL.LCModel/SIL.LCModel.csproj#L87)):*

Nice. This will make things more maintainable. 😎
</blockquote></details>

Sorry, I meant that as a compliment. Using the variable instead of hard-coding means we won't need to update this later.

I don't see a need to remove the target framework from the path; it can be useful when different artifacts target different frameworks. Although if there is a good reason to remove it from the path, I'm open to it.

josephmyers:
___
*[src/SIL.LCModel/SIL.LCModel.csproj, line 87 at r1](https://reviewable.io/reviews/sillsdev/liblcm/221#-MwbSHk8CZyqpceDTQyP:-Mx8NKzZ6Y6XYDe0IKR3:bcr7a31) ([raw file](https://github.com/sillsdev/liblcm/blob/9abb84b603635746520576d9ac710bb13a78d949/src/SIL.LCModel/SIL.LCModel.csproj#L87)):*

Sorry, I meant that as a compliment. Using the variable instead of hard-coding means we won't need to update this later.

I don't see a need to remove the target framework from the path; it can be useful when different artifacts target different frameworks. Although if there is a good reason to remove it from the path, I'm open to it.
</blockquote></details>

Yeah, I agree. I identified this as an issue when running the unit tests, with which I eventually realized that, with the .NET Standard move, output lib's were being copied to two different folders. This was causing various missing DLL problems exposed at runtime. Take a look at the corresponding commit message for more info!

josephmyers:
I'm highlighting here to me the most likely (if any) things to cause issues:
- Adding NHunspell reference in addition to NHunspell.Patched (required to compile -- unresolved symbols with .Patched only)
- Replacing Mono.Posix reference with Mono.Posix.NETStandard in Utils (only testing I did in Linux was running unit tests)
- Removing the TargetFramework from the build output path

megahirt (commented):
> Reviewed 6 of 15 files at r1, 2 of 4 files at r2, 1 of 1 files at r3, 6 of 9 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/221)* status: 16 of 19 files reviewed, 1 unresolved discussion (waiting on @megahirt and @papeh)

___
*[src/SIL.LCModel/DomainServices/WritingSystemServices.cs, line 98 at r6](https://reviewable.io/reviews/sillsdev/liblcm/221#-MxhYuclBjMPx42EY0__:-MxhYuclBjMPx42EY0_a:b-xlz3mn) ([raw file](https://github.com/sillsdev/liblcm/blob/cb2f67dc8ce604bc4c48eec7c3702297d668c9f4/src/SIL.LCModel/DomainServices/WritingSystemServices.cs#L98)):*
> ```Smalltalk
> 				var tpb = TsStringUtils.MakePropsBldr();
> 				tpb.SetIntPropValues((int)FwTextPropType.ktptForeColor, (int)FwTextPropVar.ktpvDefault,
> 					(int)ColorUtil.ConvertColorToBGR(Color.FromArgb(unchecked((int) 0xFF716F64))));
> ```

This is a curious change...

@josephmyers explained to me that `KnownColor` is not part of .Net Standard, so he needed to provide a `KnownColor.ControlDarkDark` equivalent.

ermshiperete:
> > Reviewed 6 of 15 files at r1, 2 of 4 files at r2, 1 of 1 files at r3, 6 of 9 files at r6, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/221)_ status: 16 of 19 files reviewed, 1 unresolved discussion (waiting on @megahirt and @papeh)
> 
> _[src/SIL.LCModel/DomainServices/WritingSystemServices.cs, line 98 at r6](https://reviewable.io/reviews/sillsdev/liblcm/221#-MxhYuclBjMPx42EY0__:-MxhYuclBjMPx42EY0_a:b-xlz3mn) ([raw file](https://github.com/sillsdev/liblcm/blob/cb2f67dc8ce604bc4c48eec7c3702297d668c9f4/src/SIL.LCModel/DomainServices/WritingSystemServices.cs#L98)):_
> 
> > ```smalltalk
> > 				var tpb = TsStringUtils.MakePropsBldr();
> > 				tpb.SetIntPropValues((int)FwTextPropType.ktptForeColor, (int)FwTextPropVar.ktpvDefault,
> > 					(int)ColorUtil.ConvertColorToBGR(Color.FromArgb(unchecked((int) 0xFF716F64))));
> > ```
> 
> This is a curious change...
> 
> @josephmyers explained to me that `KnownColor` is not part of .Net Standard, so he needed to provide a `KnownColor.ControlDarkDark` equivalent.

Might be good to define a const for that - it's hard to imagine what color is behind the RGB value.

ermshiperete:
Looking at the generated nuget package, I notice that the .config file now ends up in `/lib/` instead of `/lib/<framework>` (e.g. for SIL.LCModel.Core). This might not be a problem, but should be verified: create a test project, add SIL.LCModel.Core nuget package and build, then check the output directory. The results should be the same for the nuget package created from this PR (available in the Artifacts tab on AppVeyor) and the nuget package available on nuget.org.

josephmyers:
> 

Man, I have been trying to get a package installed with a local Nuget feed all day today, with no luck. Right now I'm using a .nupkg from a different PR to verify it's not an issue with my package. This is the right path to go down, correct?

image

josephmyers:
> Looking at the generated nuget package, I notice that the .config file now ends up in `/lib/` instead of `/lib/<framework>` (e.g. for SIL.LCModel.Core). This might not be a problem, but should be verified: create a test project, add SIL.LCModel.Core nuget package and build, then check the output directory. The results should be the same for the nuget package created from this PR (available in the Artifacts tab on AppVeyor) and the nuget package available on nuget.org.

Good catch, this is actually causing a problem for consumers of these packages. 
`C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(5131,5): error MSB3030: Could not copy the file "C:\Users\joseph\source\test461\test\packages\SIL.LCModel.Core.10.2\lib\net461\SIL.LCModel.Core.dll.config" because it was not found.`
What I am trying to determine now is why it's looking for the .config file in the net461 folder.

src/SIL.LCModel.Build.Tasks/SIL.LCModel.Build.Tasks.csproj
```diff
@@ -32,11 +32,11 @@ SIL.LCModel.Build.Tasks provides msbuild tasks for generating C# classes for the
     <ItemGroup>
       <None Include="$(PackageId).props" Pack="true" PackagePath="build/" />
       <None Include="$(PackageId).props" Pack="true" PackagePath="buildMultiTargeting/" />
-      <None Include="$(OutputPath)/net461/antlr.runtime.dll" Pack="true" PackagePath="tools/net461/" />
-      <None Include="$(OutputPath)/net461/IDLImporter.dll" Pack="true" PackagePath="tools/net461/" />
-      <None Include="$(OutputPath)/net461/IDLImporter.xml" Pack="true" PackagePath="tools/net461/" />
-      <None Include="$(OutputPath)/net461/Newtonsoft.Json.dll" Pack="true" PackagePath="tools/net461/" />
-      <None Include="$(OutputPath)/net461/NVelocity.dll" Pack="true" PackagePath="tools/net461/" />
+      <None Include="$(OutputPath)/antlr.runtime.dll" Pack="true" PackagePath="tools/" />
```
ermshiperete, line 35: IMO it would be better to put the files in `tools/netstandard2.0` (or whatever the correct directory name is) in the nuget package. That way it's easier to see that the files got build for netstandard2.0.

josephmyers:
With the latest changes, the .dll.config is in the /lib/ folder, but it doesn't impede the build for package consumers.

src/SIL.LCModel.Build.Tasks/SIL.LCModel.Build.Tasks.csproj
josephmyers: If I'm understanding you, that would just be changing the PackagePath for these third-party lib's back to including the target framework. This should be simple, but at first glance I wonder: do you think it will lead people to believe that those lib's were built for netstandard2.0? They may not be.

ermshiperete: > do you think it will lead people to believe that those lib's were built for netstandard2.0? They may not be.

Good point. The main dll is build for netstandard2.0 though. But that raises another interesting question: if the other libs are not build for netstandard2.0, will they work in all environments that can consume netstandard2.0? I'm thinking especially about .NET Core 3.1.

But that might not be terribly important since this dll defines msbuild tasks anyway which are then used in a process running `msbuild` or `dotnet build`. And we can probably require that the machine building that client project has a newer framework version installed, so I assume we can ignore this problem.

ermshiperete (approved):

src/SIL.LCModel.Build.Tasks/SIL.LCModel.Build.Tasks.csproj
josephmyers: In short, I don't know. Per [this link](https://stackoverflow.com/a/46394435), it is quite possible, but it depends totally on what calls are being made in the Framework lib's and whether Core 3.1 API supports/provides those calls or not.

jasonleenaylor (changes requested):
> Reviewed 2 of 9 files at r6, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/221)* status: all files reviewed, 3 unresolved discussions (waiting on @ermshiperete and @josephmyers)

___
*[src/SIL.LCModel/DomainServices/WritingSystemServices.cs, line 97 at r9](https://reviewable.io/reviews/sillsdev/liblcm/221#-My7xO0eDbegtxx8OVfe:-My7xO0f--e2aba0DpBL:b-e8t4lv) ([raw file](https://github.com/sillsdev/liblcm/blob/46463ebae0d440abc247b9e3a3cc2d47c8d49d48/src/SIL.LCModel/DomainServices/WritingSystemServices.cs#L97)):*
> ```Smalltalk
> 			{
> 				var tpb = TsStringUtils.MakePropsBldr();
> 				const int darkColor = unchecked((int) 0xFF716F64);
> ```

We aren't doing any math here, why is unchecked necessary?



_Code quote:_
```Smalltalk
unchecked
```
___
*[src/SIL.LCModel.Core/SIL.LCModel.Core.csproj, line 16 at r9](https://reviewable.io/reviews/sillsdev/liblcm/221#-My7y6aLE433Ka6rSkMA:-My7y6aM3onL_TI3iK3K:byrcltj) ([raw file](https://github.com/sillsdev/liblcm/blob/46463ebae0d440abc247b9e3a3cc2d47c8d49d48/src/SIL.LCModel.Core/SIL.LCModel.Core.csproj#L16)):*
> ```XML
>     <PackageReference Include="Icu4c.Win.Fw.Lib" Version="70.1.47" IncludeAssets="build" PrivateAssets="runtime" />
>     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
>     <PackageReference Include="NHunspell" Version="1.2.5554.16953" />
> ```

So I hadn't responded earlier because I was looking into this. I couldn't understand why both would be needed. Digging around it looks like @gtryus is probably the person who published NHunspell.Patched. However, I can't see any substantive changes in https://github.com/sillsdev/CompactNHunspell that explain why we needed an NHunspell.Patched package to begin with. 
I don't want to see both versions here. It is a code smell, and is likely to bite us in the future.



_Code quote:_
```XML
<PackageReference Include="NHunspell" Version="1.2.5554.16953" />
```

gtryus:
I don't remember making any patched version of NHunspell...

josephmyers:
> _[src/SIL.LCModel/DomainServices/WritingSystemServices.cs, line 97 at r9](https://reviewable.io/reviews/sillsdev/liblcm/221#-My7xO0eDbegtxx8OVfe:-My7xO0f--e2aba0DpBL:b-e8t4lv) ([raw file](https://github.com/sillsdev/liblcm/blob/46463ebae0d440abc247b9e3a3cc2d47c8d49d48/src/SIL.LCModel/DomainServices/WritingSystemServices.cs#L97)):_
> 
> > ```smalltalk
> > 			{
> > 				var tpb = TsStringUtils.MakePropsBldr();
> > 				const int darkColor = unchecked((int) 0xFF716F64);
> > ```
> 
> We aren't doing any math here, why is unchecked necessary?
> 

It wouldn't compile without that. I was a little confused, as well, but it probably has to do with warnings about potential overflows. I didn't consider it worth spending a ton of time investigating, but maybe I'm wrong?

josephmyers:
> So I hadn't responded earlier because I was looking into this. I couldn't understand why both would be needed. Digging around it looks like @gtryus is probably the person who published NHunspell.Patched. However, I can't see any substantive changes in https://github.com/sillsdev/CompactNHunspell that explain why we needed an NHunspell.Patched package to begin with. I don't want to see both versions here. It is a code smell, and is likely to bite us in the future.
> 
> _Code quote:_
> 
> ```
> <PackageReference Include="NHunspell" Version="1.2.5554.16953" />
> ```

[I was unable](https://github.com/sillsdev/liblcm/pull/221/commits/dee97c97d639a44b4f605f50cfacce04767524b0) to get it to compile with the original Patched, but it was necessary to retain the Patched to keep the tests passing. IMHO, the root code smell is the custom .Patched packages themselves, and I am very in favor of removing this one if we no longer need those Hunspellx86 or Hunspellx64 lib's. In general, I was trying to preserve as much of the existing functionality as possible, having no knowledge of how old/unused any piece is.

josephmyers:
___
*[src/SIL.LCModel.Core/SIL.LCModel.Core.csproj, line 16 at r9](https://reviewable.io/reviews/sillsdev/liblcm/221#-My7y6aLE433Ka6rSkMA:-MyBN5ZtEFnFYzhbCKEQ:b-896fix) ([raw file](https://github.com/sillsdev/liblcm/blob/46463ebae0d440abc247b9e3a3cc2d47c8d49d48/src/SIL.LCModel.Core/SIL.LCModel.Core.csproj#L16)):*

So I hadn't responded earlier because I was looking into this. I couldn't understand why both would be needed. Digging around it looks like @gtryus is probably the person who published NHunspell.Patched. However, I can't see any substantive changes in https://github.com/sillsdev/CompactNHunspell that explain why we needed an NHunspell.Patched package to begin with. 
I don't want to see both versions here. It is a code smell, and is likely to bite us in the future.
</blockquote></details>

Done.

josephmyers:
___
*[src/SIL.LCModel/DomainServices/WritingSystemServices.cs, line 97 at r9](https://reviewable.io/reviews/sillsdev/liblcm/221#-My7xO0eDbegtxx8OVfe:-MyFYWes8C93y2Zkvsmf:bb6un1j) ([raw file](https://github.com/sillsdev/liblcm/blob/46463ebae0d440abc247b9e3a3cc2d47c8d49d48/src/SIL.LCModel/DomainServices/WritingSystemServices.cs#L97)):*

Hmm, it is probably fine, but even so using the color this way does still obscure things.
How about:
var darkColor = Color.FromArgb(64,64,64);
</blockquote></details>

Done

jasonleenaylor (commented):
I'm happy with this, we just need to do an integration test with FieldWorks now. 

> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/221#-:-MyFsXzK619a698uVoYN:b-rd54jr)* status: all files reviewed, 1 unresolved discussion (waiting on @ermshiperete)

megahirt:
Thanks Jason!

josephmyers (commented):
> Reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/221)* status: 19 of 21 files reviewed, 1 unresolved discussion (waiting on @ermshiperete and @jasonleenaylor)

josephmyers (commented):
> Reviewed 5 of 15 files at r1, 3 of 4 files at r2, 1 of 1 files at r3, 4 of 9 files at r6, 1 of 1 files at r8, 1 of 1 files at r11, 2 of 2 files at r12.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/221)* status: all files reviewed, 1 unresolved discussion (waiting on @ermshiperete and @jasonleenaylor)

PR #246, opened 2022-08-22
LT-20698: Use non-cultured date formats

papeh:
Write dates using a culture-neutral format ensures they are consistent and anyone can read them.

TODO: test

Part of https://jira.sil.org/browse/LT-20698

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/246)


papeh:
Only user-facing strings are using culture-sensitive date formats. (Perhaps a code comment is in order)

PR #270, opened 2023-01-05
use fallback for MemoryMappedFile.CreateOrOpen when not running on windows

hahn-kev:
Creating this PR to fix an issue seen here: https://github.com/sillsdev/LfMerge/issues/317

open to suggestions, maybe remove the `IsMono` check? But that might cause other issues, possibly for mono running on windows.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/270)

### DIFF
--- a/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs
+++ b/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs
@@ -262,7 +262,7 @@ namespace SIL.LCModel.Infrastructure.Impl
 
 		private MemoryMappedFile CreateOrOpen(string name, long capacity, bool createdNew)
 		{
-			if (MiscUtils.IsMono || !MiscUtils.IsWindows)
+			if (SIL.PlatformUtilities.Platform.IsUnix)
 			{
 				name = Path.Combine(m_commitLogDir, name);
 				// delete old file that could be left after a crash
--- a/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs
+++ b/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs
@@ -262,7 +262,7 @@ namespace SIL.LCModel.Infrastructure.Impl
 
 		private MemoryMappedFile CreateOrOpen(string name, long capacity, bool createdNew)
 		{
-			if (MiscUtils.IsMono)
+			if (MiscUtils.IsMono || !MiscUtils.IsWindows)
 			{
 				name = Path.Combine(m_commitLogDir, name);
 				// delete old file that could be left after a crash
--- a/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs
+++ b/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs
@@ -262,7 +262,7 @@ namespace SIL.LCModel.Infrastructure.Impl
 
 		private MemoryMappedFile CreateOrOpen(string name, long capacity, bool createdNew)
 		{
-			if (SIL.PlatformUtilities.Platform.IsUnix)
+			if (MiscUtils.IsUnix)
 			{
 				name = Path.Combine(m_commitLogDir, name);
 				// delete old file that could be left after a crash
--- a/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs
+++ b/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2013-2014 SIL International
+// Copyright (c) 2013-2023 SIL International
 // This software is licensed under the LGPL, version 2.1 or later
 // (http://www.gnu.org/licenses/lgpl-2.1.html)
 
@@ -11,6 +11,7 @@ using System.Linq;
 using ProtoBuf;
 using SIL.LCModel.DomainServices.DataMigration;
 using SIL.LCModel.Utils;
+using SIL.PlatformUtilities;
 using SIL.Threading;
 
 namespace SIL.LCModel.Infrastructure.Impl
@@ -43,7 +44,7 @@ namespace SIL.LCModel.Infrastructure.Impl
 		{
 			m_peerProcesses = new Dictionary<int, Process>();
 			m_peerID = Guid.NewGuid();
-			if (MiscUtils.IsMono)
+			if (Platform.IsMono)
 			{
 				// /dev/shm is not guaranteed to be available on all systems, so fall back to temp
 				m_commitLogDir = Directory.Exists("/dev/shm") ? "/dev/shm" : Path.GetTempPath();
@@ -156,7 +157,7 @@ namespace SIL.LCModel.Infrastructure.Impl
 								metadata.FileGeneration = metadata.CurrentGeneration;
 							}
 							RemovePeer(metadata, m_peerID);
-							delete = MiscUtils.IsMono && metadata.Peers.Count == 0;
+							delete = Platform.IsMono && metadata.Peers.Count == 0;
 							SaveMetadata(stream, metadata);
 						}
 					}
@@ -262,14 +263,14 @@ namespace SIL.LCModel.Infrastructure.Impl
 
 		private MemoryMappedFile CreateOrOpen(string name, long capacity, bool createdNew)
 		{
-			if (MiscUtils.IsUnix)
+			if (Platform.IsUnix)
 			{
 				name = Path.Combine(m_commitLogDir, name);
 				// delete old file that could be left after a crash
 				if (createdNew && File.Exists(name))
 					File.Delete(name);
 
-				// Mono only supports memory mapped files that are backed by an actual file
+				// Unix supports memory-mapped files only if they are backed by an actual file
 				if (!File.Exists(name))
 				{
 					using (var fs = new FileStream(name, FileMode.CreateNew))


megahirt:
@hahn-kev and I looked at this together in response to two users whose projects fail to S/R on language forge because of an exception in liblcm.

I am in favor of changing the if statement to simply 
```
if (!MiscUtils.IsWindows)
```
since running Mono on Windows isn't an use case for liblcm as far as I understand.

megahirt:
Looking at this code has me wondering if we can simply always use the platform agnostic code path of `MemoryMappedFile.CreateFromFile(name)` in all cases including Windows, but I don't understand if that would change behavior in any way on Windows.

src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs
```diff
@@ -262,7 +262,7 @@ private void CreateSharedMemory(bool createdNew)
 
 		private MemoryMappedFile CreateOrOpen(string name, long capacity, bool createdNew)
 		{
-			if (MiscUtils.IsMono)
+			if (MiscUtils.IsMono || !MiscUtils.IsWindows)
```
papeh, line 265: Would `SIL.PlatformUtilities.Platform.IsUnix` (alone) work here? If not, `!Platform.IsWindows` alone should work. Both are in `SIL.Core`. Either of these :lgtm:

papeh (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/270)* status: all files reviewed, 2 unresolved discussions (waiting on @hahn-kev)

___
*[`src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs` line 265 at r1](https://reviewable.io/reviews/sillsdev/liblcm/270#-NL03aDY5i04QcUmxrUc-r1-265:-NL1PG4W-USWPtPYAh41:b-l7n4kb) ([raw file](https://github.com/sillsdev/liblcm/blob/2090c0328e53dc438cb15b837d8d758866befffa/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs#L265)):*
<details><summary><i>Previously, hahn-kev (Kevin Hahn) wrote…</i></summary><blockquote>

I'm not sure, I didn't know if Unix would work for both Linux and Mac (or something else). However the way the failing API is written is that it only works for Windows, so this seemed the way to go instead of something else as the API is only supported in windows, rather than not being supported in Unix. Though that may end up being the same thing at the end of the day. 
</blockquote></details>

When this was originally written, Mono, Linux, and Unix were effectively synonymous, and Windows was the only other choice. If you think checking for Windows instead of Unix would be more future-proof, we can check for Windows instead.
___
*[`src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs` line 272 at r1](https://reviewable.io/reviews/sillsdev/liblcm/270#-NL1Ot4H5CZW4KER0ENu:-NL1Ot4H5CZW4KER0ENv:b-t8swzm) ([raw file](https://github.com/sillsdev/liblcm/blob/2090c0328e53dc438cb15b837d8d758866befffa/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs#L272)):*
> ```Smalltalk
> 					File.Delete(name);
> 
> 				// Mono only supports memory mapped files that are backed by an actual file
> ```

Per the discussion, this seems to be a problem with Unix rather than Mono



_Suggestion:_
```Smalltalk
Unix 
```

megahirt (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/270)* status: all files reviewed, 2 unresolved discussions (waiting on @papeh)

___
*[`src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs` line 265 at r1](https://reviewable.io/reviews/sillsdev/liblcm/270#-NL03aDY5i04QcUmxrUc-r1-265:-NL1ju4G07f96Hx83QVd:b-xuklt0) ([raw file](https://github.com/sillsdev/liblcm/blob/2090c0328e53dc438cb15b837d8d758866befffa/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs#L265)):*

When this was originally written, Mono, Linux, and Unix were effectively synonymous, and Windows was the only other choice. If you think checking for Windows instead of Unix would be more future-proof, we can check for Windows instead.
</blockquote></details>

If we have to keep an if statement, then I would prefer
 `if (!SIL.PlatformUtilities.Platform.IsWindows)`

@papeh what do you think about ditching the named memory mapped files entirely (a windows-specific feature) and go with the implementation as coded on all platforms, including windows?  I know we'd have to test it somehow (does the program run?) but I think it would be reduced complexity to simply do the same on all platforms.
___
*[`src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs` line 272 at r1](https://reviewable.io/reviews/sillsdev/liblcm/270#-NL1Ot4H5CZW4KER0ENu:-NL1kbnDBV7HkxRZeL1j:bjrr915) ([raw file](https://github.com/sillsdev/liblcm/blob/2090c0328e53dc438cb15b837d8d758866befffa/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs#L272)):*

Per the discussion, this seems to be a problem with Unix rather than Mono
</blockquote></details>

It is a issue with Unix/Linux systems, not Mono per say, but the comment is basically referring to the only use-case which is Mono on Linux.

papeh (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/270)* status: all files reviewed, 2 unresolved discussions (waiting on @hahn-kev and @megahirt)

___
*[`src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs` line 265 at r1](https://reviewable.io/reviews/sillsdev/liblcm/270#-NL03aDY5i04QcUmxrUc-r1-265:-NL1mRcv3-LFN_gf6rkg:b1b7t1k) ([raw file](https://github.com/sillsdev/liblcm/blob/2090c0328e53dc438cb15b837d8d758866befffa/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs#L265)):*
<details><summary><i>Previously, megahirt (Christopher Hirt) wrote…</i></summary><blockquote>

If we have to keep an if statement, then I would prefer
 `if (!SIL.PlatformUtilities.Platform.IsWindows)`

@papeh what do you think about ditching the named memory mapped files entirely (a windows-specific feature) and go with the implementation as coded on all platforms, including windows?  I know we'd have to test it somehow (does the program run?) but I think it would be reduced complexity to simply do the same on all platforms.
</blockquote></details>

That sounds reasonable.

papeh (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/270)* status: all files reviewed, 2 unresolved discussions (waiting on @hahn-kev and @megahirt)

___
*[`src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs` line 265 at r1](https://reviewable.io/reviews/sillsdev/liblcm/270#-NL03aDY5i04QcUmxrUc-r1-265:-NL1mmmp5CAma4xRskvo:bkmkw1n) ([raw file](https://github.com/sillsdev/liblcm/blob/2090c0328e53dc438cb15b837d8d758866befffa/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs#L265)):*

That sounds reasonable.
</blockquote></details>

Although I wonder how the file-backed file would affect performance.

megahirt (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/270)* status: all files reviewed, 2 unresolved discussions (waiting on @jasonleenaylor and @papeh)

___
*[`src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs` line 265 at r1](https://reviewable.io/reviews/sillsdev/liblcm/270#-NL03aDY5i04QcUmxrUc-r1-265:-NL5kpTyEQkzfi3HxeBi:bm3mdxw) ([raw file](https://github.com/sillsdev/liblcm/blob/2090c0328e53dc438cb15b837d8d758866befffa/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs#L265)):*

Although I wonder how the file-backed file would affect performance.
</blockquote></details>

Yeah I wondered about that too, however I realistically don't know how to test the performance, or even trigger this code path (could take some guesses).  @jasonleenaylor Do you have an opinion here?

Just from the class name `MemoryMappedFile` I assume we are always dealing with the filesystem, even if Windows has an extra api method that abstracts the file into just a name without a path.  So that makes me think the Windows-specific method is doing something similar to what we are doing in our code.  Can we look at the source of how `MemoryMappedFile.CreateOrOpen(name, capacity)` is implemented to see if it's similar to our own implementation?

papeh (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/270)* status: all files reviewed, 2 unresolved discussions (waiting on @jasonleenaylor)

___
*[`src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs` line 265 at r1](https://reviewable.io/reviews/sillsdev/liblcm/270#-NL03aDY5i04QcUmxrUc-r1-265:-NLNi-vxCCAaN5gAvYHS:b-v3lkzz) ([raw file](https://github.com/sillsdev/liblcm/blob/2090c0328e53dc438cb15b837d8d758866befffa/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs#L265)):*
<details><summary><i>Previously, megahirt (Christopher Hirt) wrote…</i></summary><blockquote>

Yeah I wondered about that too, however I realistically don't know how to test the performance, or even trigger this code path (could take some guesses).  @jasonleenaylor Do you have an opinion here?

Just from the class name `MemoryMappedFile` I assume we are always dealing with the filesystem, even if Windows has an extra api method that abstracts the file into just a name without a path.  So that makes me think the Windows-specific method is doing something similar to what we are doing in our code.  Can we look at the source of how `MemoryMappedFile.CreateOrOpen(name, capacity)` is implemented to see if it's similar to our own implementation?
</blockquote></details>

Resharper VS plugin would allow viewing of decompiled MS sources. If neither of you have a license, I could take a look sometime.

hahn-kev:
it looks like the rough logic of the windows API is to just create the mapped file, and if that fails try to open one instead. If that fails then it tries again 14 times, if it still fails then it throws an error. Otherwise it'll return.

Our current logic basically deletes the file (if it exists and creating a new one was requested), then if the file doesn't exist (because we deleted it or it never existed) we create it , then just return the memory mapped file.

The biggest difference that I see (other then the whole retry functionality) is that we will delete an old file if we want to create a new one, where as the windows API will never do that. It's not clear to me why the API can't be ported to work on other platforms so there's probably something I'm missing with my understanding of the APIs.

ermshiperete:
In my understanding, on Windows a memory mapped file can be [persisted or non-persisted](https://learn.microsoft.com/en-us/dotnet/standard/io/memory-mapped-files). Only the former type uses a file on disk, in the latter case it's just shared memory. The .NET implementation directly uses the Windows system APIs. On Linux a memory mapped file always has a corresponding file on disk, IIRC. For the non-persistent type you'd have to implement it with a different system API. With that API it might be way harder to implement the methods of `MemoryMappedFile`.

megahirt:
> In my understanding, on Windows a memory mapped file can be [persisted or non-persisted](https://learn.microsoft.com/en-us/dotnet/standard/io/memory-mapped-files). Only the former type uses a file on disk, in the latter case it's just shared memory. The .NET implementation directly uses the Windows system APIs. On Linux a memory mapped file always has a corresponding file on disk, IIRC. For the non-persistent type you'd have to implement it with a different system API. With that API it might be way harder to implement the functionality of `MemoryMappedFile`.

Thanks for that, Eberhard.  In your opinion, would it be worth standardizing on the same implementation (persistent memory-mapped file) across all platforms?  Is there somehow a performance hit to using a file-based memory mapped file?

jasonleenaylor (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/270)* status: all files reviewed, 2 unresolved discussions (waiting on @papeh)

___
*[`src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs` line 265 at r1](https://reviewable.io/reviews/sillsdev/liblcm/270#-NL03aDY5i04QcUmxrUc-r1-265:-NLRwj2E3EjrTbvCaVu9:b-nm6vsj) ([raw file](https://github.com/sillsdev/liblcm/blob/2090c0328e53dc438cb15b837d8d758866befffa/src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs#L265)):*

Resharper VS plugin would allow viewing of decompiled MS sources. If neither of you have a license, I could take a look sometime.
</blockquote></details>

I would prefer that we leave the windows implementation alone. The SharedXMLBackendProvider is used as the storage for the project when the project settings specify it, or if there is Scripture content present in the project. I added a comment about this into the LFMerge issue. Platform.IsUnix is appropriate for this code, it should work on all Unix like platforms..

ermshiperete:
>  Is there somehow a performance hit to using a file-based memory mapped file?

My gut feeling is that it would have a performance impact since it hits the file system, but I have no idea how much of a difference it would make.

I'd agree with @jasonleenaylor that we shouldn't change the Windows implementation.

src/SIL.LCModel/Infrastructure/Impl/SharedXMLBackendProvider.cs
megahirt, line 265: Based on further conversation with @jasonleenaylor and @ermshiperete we will leave the Windows implementation alone and stick with a simple check for "isUnix":
```suggestion
			if (SIL.PlatformUtilities.Platform.IsUnix)
```

megahirt (changes requested):
@hahn-kev Let's get this line changed (suggestion provided) and then get this merged so that we can attempt to move this toward a user fix.

megahirt:
> I'd agree with @jasonleenaylor that we shouldn't change the Windows implementation.

Thank you Jason and Eberhard for weighing in on this.  We'll go with the smallest possible change as you suggested and leave the Windows implementation alone.

megahirt (approved):
Thanks @hahn-kev

megahirt (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/270)* status: all files reviewed, 2 unresolved discussions (waiting on @ermshiperete and @papeh)

ermshiperete (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/270)* status: all files reviewed, 2 unresolved discussions (waiting on @papeh)

papeh (approved):

papeh (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/270)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @hahn-kev)

PR #272, opened 2023-01-16
use fallback for MemoryMappedFile.CreateOrOpen when not running on windows

hahn-kev:
previous pr #270 was merged into master instead of develop as it should have been.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/272)

PR #280, opened 2023-04-21
Fix NullReferenceError in SharedXMLBackendProvider under .Net Core on Linux

rmunn:
PR #272 changed `Platform.IsMono` to `IsUnix` in the `CreateOrOpen` method of SharedXMLBackendProvider, but the `Platform.IsMono` check in the constructor was missed. This fixes that omission, and will also fix https://github.com/sillsdev/LfMerge/issues/323 (which was caused by the mismatch: when liblcm is running under .Net Core on Linux, as in LfMerge, the CreateOrOpen method ends up trying to use the `m_commitLogDir` property that the constructor never set).


megahirt:
I asked @jasonleenaylor about this a while back and he insightfully pointed out that there is no point in LFMerge even invoking SharedXMLBackendProvider since it doesn't make sense in the server context, and that there was a way to tell liblcm to "turn if off" in a setting somewhere, but I don't know how to do that.  So LFMerge is exercising this code when it's not actually useful to our purposes on the server.  That's my understanding.

PR #291, opened 2023-12-14
Fix undo problem when lowercase matching is used

mark-sil:
Undo was not working when an uppercase word used a guess for the lowercase word.  The undo change was getting added for the lowercase word. This pr fixes the problem by adding a undo change for the uppercase word.
https://jira.sil.org/browse/LT-21424

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/291)


mark-sil:
Decided to close this PR after discussion with Jason.  Looking for a different solution that doesn’t add another change.

PR #295, opened 2024-04-24
LT-21728 (prioritize parses) and LT-12633 (always suggest lexical entries regardless of case)

jtmaxwell3:
- Add GetSortedAnalysisGuesses, so that the user menu of analyses can be sorted by likelihood.
- Add GetSortedGlossGuesses, so that the user menu of glosses can be sorted by likelihood.
- Change the likelihood calculation for GetBestGuess and GetSortedxxxGuesses to consider the previous word.
- Build the GuessServices cache incrementally instead of all at once to improve performance with large corpora.
- Add lowercase analyses to candidates for GetBestGuess and GetSortedAnalysesGuesses for relevant uppercase forms.
- Sort uppercase and lowercase analyses together using the same likelihood calculation.
- Prefer uppercase analyses over lowercase analyses all other things are equal.
- Generate lexical entries for both uppercase and lowercase forms when relevant (LT-12633).

Fieldworks will need to be updated to call GetSortedAnalysisGuesses and GetSortedGlossGuesses.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/295)

### DIFF
--- a/src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs
+++ b/src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs
@@ -9,12 +9,14 @@
 // </remarks>
 
 using System.Collections.Generic;
+using System.Configuration;
 using System.Linq;
 using Icu;
 using SIL.LCModel.Core.KernelInterfaces;
 using SIL.LCModel.Core.Text;
 using SIL.LCModel.Core.WritingSystems;
 using SIL.LCModel.Infrastructure;
+
 
 namespace SIL.LCModel.DomainServices
 {
@@ -32,6 +34,10 @@ namespace SIL.LCModel.DomainServices
 		public AnalysisGuessServices(LcmCache cache)
 		{
 			Cache = cache;
+			var wfFactory = Cache.ServiceLocator.GetInstance<IWfiWordformFactory>();
+			var wsVern = Cache.DefaultVernWs;
+			m_emptyWAG = wfFactory.Create(TsStringUtils.MakeString("", wsVern));
+			m_nullWAG = new NullWAG();
 		}
 
 		/// <summary>
@@ -100,17 +106,27 @@ namespace SIL.LCModel.DomainServices
 			set { m_parserApprovedTable = value; }
 		}
 
+		class PriorityCount
+		{
+			public int priority = 0;
+			public int count = 0;
+		}
+
 		private IDictionary<IAnalysis, IAnalysis> m_guessTable;
 		IDictionary<IAnalysis, IAnalysis> GuessTable
 		{
 			get
 			{
 				if (m_guessTable == null)
-					LoadGuessTable();
+					// LoadGuessTable();
+					GuessTable = new Dictionary<IAnalysis, IAnalysis>();
 				return m_guessTable;
 			}
 			set { m_guessTable = value; }
 		}
+
+		private readonly IAnalysis m_emptyWAG;
+		private readonly IAnalysis m_nullWAG;
 
 		/// <summary>
 		/// Informs the guess service that the indicated occurrence is being replaced with the specified new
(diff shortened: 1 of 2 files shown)


jasonleenaylor (commented):
> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/295)* status: 1 of 2 files reviewed, all discussions resolved

___
*[`src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs` line 64 at r1](https://reviewable.io/reviews/sillsdev/liblcm/295#-NwGho5PDhVDtinP_WLc:-NwGho5PDhVDtinP_WLd:bd56cuz) ([raw file](https://github.com/sillsdev/liblcm/blob/108c08c9ae2132dcb8a9b7313c6c264ddad46b2f/src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs#L64)):*
> ```Smalltalk
> 		LcmCache Cache { get; set; }
> 
> 		class PriorityCount
> ```

A class comment here would be welcome.
___
*[`src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs` line 88 at r1](https://reviewable.io/reviews/sillsdev/liblcm/295#-NwGlVZOBNQm2TvYd3UZ:-NwGlVZOBNQm2TvYd3U_:b-jbga2f) ([raw file](https://github.com/sillsdev/liblcm/blob/108c08c9ae2132dcb8a9b7313c6c264ddad46b2f/src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs#L88)):*
> ```Smalltalk
> 		// GuessTable2 is like GuessTable, but for uppercase word forms that can have lowercase analyses.
> 		private IDictionary<IAnalysis, Dictionary<IAnalysis, IAnalysis>> m_guessTable2;
> 		IDictionary<IAnalysis, Dictionary<IAnalysis, IAnalysis>> GuessTable2
> ```

I think we can improve this name, but I'm trying to find one that isn't absurdly long.
CaseInsensitiveGuessTable
LowerCaseGuessTable
GuessTableWithLower
GuessTableLower
I think I can live with GuessTableLower if that doesn't seem likely to be misleading.
___
*[`src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs` line 633 at r1](https://reviewable.io/reviews/sillsdev/liblcm/295#-NwGn97H4k4meSdujMKP:-NwGn97H4k4meSdujMKQ:b-lm3bsb) ([raw file](https://github.com/sillsdev/liblcm/blob/108c08c9ae2132dcb8a9b7313c6c264ddad46b2f/src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs#L633)):*
> ```Smalltalk
> 		{
> 			// first see if there is a relevant lowercase form of a sentence initial (non-lowercase) wordform
> 			// TODO: make it look for the first word in the sentence...may not be at Index 0!
> ```

True.  I feel like there is already a method that returns the first word in a segment...
___
*[`src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs` line 713 at r1](https://reviewable.io/reviews/sillsdev/liblcm/295#-NwGnlop41gMWNd5X2Nt:-NwGnlop41gMWNd5X2Nu:b-5i67g8) ([raw file](https://github.com/sillsdev/liblcm/blob/108c08c9ae2132dcb8a9b7313c6c264ddad46b2f/src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs#L713)):*
> ```Smalltalk
> 		public List<IWfiAnalysis> GetSortedAnalysisGuesses(IWfiWordform wordform, AnalysisOccurrence occurrence = null, bool onlyIndexZeroLowercaseMatching = true)
> 		{
> 			// Get the writing system.  Should this be passed in as an argument?
> ```

If it is available in most callers then I would make it an argument, otherwise this is fine.

jtmaxwell3:
1) I will add a comment to PriorityCount.
2) How about CaselessGuessTable, as a shortened form of CaseInsensitiveGuessTable?
3) We would need to change the parameter name "onlyIndexZeroLowercaseMatching" to "onlyFirstWordLowercaseMatching".  Wouldn't that require other users of liblcm to change their code if they use the parameter name?
4) I will add GetSortedAnalysisGuesses(IWfiWordform wordform, int ws) for cases where the client doesn't have an occurrence.

jtmaxwell3:
Good catch!  I renamed condition to context in the test file.  I also added a unit test for generating both uppercase and lowercase lexical entries for sentence-initial uppercase words.

jasonleenaylor (approved):
> Reviewed 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/295)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

PR #296, opened 2024-05-10
Lt 21425: Caps issue with parser

jtmaxwell3:
Changes the guesser to treat capitalization the same way that the parser will treat capitalization.
In particular, uppercased words are analyzed as having lowercase wordforms.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/296)

### DIFF
--- a/src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs
+++ b/src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs
@@ -636,21 +636,22 @@ namespace SIL.LCModel.DomainServices
 			// first see if there is a relevant lowercase form of a sentence initial (non-lowercase) wordform
 			// TODO: make it look for the first word in the sentence...may not be at Index 0!
 			IWfiWordform lowercaseWf = null;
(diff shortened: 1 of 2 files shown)


jasonleenaylor:
*[`src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs` line 707 at r1](https://reviewable.io/reviews/sillsdev/liblcm/296#-Nxmj3x0EguIKVO6b0lJ:-Nxmj3x0EguIKVO6b0lK:b-bhml6u) ([raw file](https://github.com/sillsdev/liblcm/blob/51ef36ca9ae936afa761b425870975896565beae/src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs#L707)):*
> ```Smalltalk
> 			if (Cache.ServiceLocator.GetInstance<IWfiWordformRepository>().TryGetObject(tssWord, out wf))
> 				return wf;
> 			// Only create a lowercase wordform if there is an entry for it in the lexicon.
> ```

So this comment is only true given the current use of the method. Should we move the lowercasing into this method?

PR #298, opened 2024-06-12
Fix crash encountered while bulk deleting entries

jasonleenaylor:
* Discovered while trying to prune entries in the project attached to https://jira.sil.org/browse/LT-18577

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/298)


jtmaxwell3 (approved):
You misspelled "yeild".

PR #306, opened 2024-07-30
run unit tests in CI against dotnet 8

hahn-kev:
Our team is planning to use LCM on dotnet 8, so I decided it would be good to run tests against dotnet 8. In doing so I found a few issues with some paths which would cause errors during build or clean.

Prevously RhinoMocks was being used however it does not support modern dotnet, Moq is the suggested replacement and I found a library that works as a shim, so I didn't need to rewrite any of the test code (with one exception).

Lastly I had to make Core and TestHelper target net8.0 also as tests would not get the desired IcuData and other dependancies otherwise.

Actual code changes:
* FileUtils.AreFilesIdentical now always uses SHA1 as the hash algorithm as modern dotnet throws if you try to delegate to a default algorithm and SHA1 is what was being used in framework
* AssertValidFilePath and IsFilePathValid had to be rewritten since modern dotnet does not do much if any validation anymore.

There's currently some failing tests due to [NHunspell](https://www.nuget.org/packages/NHunspell) not supporting anything but .Net framework and it's last update was 2015. I suggest we go to https://github.com/aarondandy/WeCantSpell.Hunspell as it's been updated in the last year and supporst .net standard 2.0. However I think that's out of scope for this PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/306)

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -67,11 +67,11 @@ jobs:
     - name: Test on Linux
       run: |
         . environ
-        dotnet test --no-restore --no-build -p:ParallelizeAssembly=false --configuration Release --logger:"trx" --results-directory ./test-results
+        dotnet test --no-restore --no-build -p:ParallelizeAssembly=false --configuration Release --logger:"trx;LogFilePrefix=results" --results-directory ./test-results
       if: matrix.os == 'ubuntu-latest'
 
     - name: Test on Windows
-      run: dotnet test --no-restore --no-build -p:ParallelizeAssembly=false --configuration Release --logger:"trx" --results-directory ./test-results
+      run: dotnet test --no-restore --no-build -p:ParallelizeAssembly=false --configuration Release --logger:"trx;LogFilePrefix=results" --results-directory ./test-results
       if: matrix.os != 'ubuntu-latest'
     - name: Upload test results
       if: always()
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -28,9 +28,9 @@ jobs:
 
     # Install the .NET Core workload
     - name: Install .NET Core
-      uses: actions/setup-dotnet@v2
+      uses: actions/setup-dotnet@v4
       with:
-        dotnet-version: 6.0.x
+        dotnet-version: 8.0.x
 
     - name: Install mono-devel package
       run: sudo apt-get install mono-devel
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -94,7 +94,7 @@ jobs:
     - name: Publish Artifacts
       uses: actions/upload-artifact@v4
       with:
-        name: NugetPackages
+        name: NugetPackages (${{ matrix.os }})
         path: artifacts/*.nupkg
       if: github.event_name == 'pull_request'
   publish-test-results:
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -11,7 +11,8 @@
     <PackageProjectUrl>https://github.com/sillsdev/liblcm</PackageProjectUrl>
     <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
     <Platform>Any CPU</Platform>
-    <OutputPath>$(MSBuildThisFileDirectory)/artifacts/$(Configuration)/$(TargetFramework)</OutputPath>
+	<Configuration Condition="'$(Configuration)' == ''">Debug</Configuration>
+    <OutputPath>$(MSBuildThisFileDirectory)artifacts/$(Configuration)/$(TargetFramework)</OutputPath>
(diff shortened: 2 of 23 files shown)


hahn-kev:
@jasonleenaylor `CollatorSort_DoesNotThrow` is currently throwing in our tests. It looks like it's due to a missmatch between the version of icu.net that SIL.WritingSystems is using and the version we're using. If I were to upgrade SIL.WritingSystems to use icu.net 2.10 would we be able to upgrade to that new WritingSystems package to fix this?

jasonleenaylor:
@hahn-kev yes, we can do that. I will want to get some other changes merged first so I can delay the corresponding work in FieldWorks until the next sprint.

jasonleenaylor (commented):
> Reviewed 3 of 11 files at r1, 3 of 12 files at r2, 7 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/306)* status: 13 of 23 files reviewed, all discussions resolved

___
*[`src/SIL.LCModel.Utils/SIL.LCModel.Utils.csproj` line 13 at r3](https://reviewable.io/reviews/sillsdev/liblcm/306#-O4MHuco6iIwpDsv3g1r:-O4MHucp9TqYgNNTwv9j:b2kr5l8) ([raw file](https://github.com/sillsdev/liblcm/blob/f4232a846d90564fff0e42d4df9cca7fbfbcc53b/src/SIL.LCModel.Utils/SIL.LCModel.Utils.csproj#L13)):*
> ```XML
>     <PackageReference Include="GitVersion.MsBuild" Version="5.6.10" PrivateAssets="All" />
>     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
>     <PackageReference Include="SIL.Core" Version="14.2.0-beta0021" />
> ```

Here, and elsewhere I'd like to keep the wildcard, I accept the risk of a breaking library change in the same version over constant update maintenance churn. Basically, I trust me/us on our core libraries.
___
*[`tests/SIL.LCModel.Tests/RhinoMocksToMoq/Expect.cs` line 1 at r3](https://reviewable.io/reviews/sillsdev/liblcm/306#-O4MJmtVA5TtrU6fTvo_:-O4MJmtVA5TtrU6fTvoa:bnzjsaz) ([raw file](https://github.com/sillsdev/liblcm/blob/f4232a846d90564fff0e42d4df9cca7fbfbcc53b/tests/SIL.LCModel.Tests/RhinoMocksToMoq/Expect.cs#L1)):*
> ```Smalltalk
> using System;
> ```

:vomiting_face: 
We should discuss, this may be the best solution but I prefer to avoid inlining 3rd party code.

hahn-kev:
*[`tests/SIL.LCModel.Tests/RhinoMocksToMoq/Expect.cs` line 1 at r3](https://reviewable.io/reviews/sillsdev/liblcm/306#-O4MJmtVA5TtrU6fTvo_:-O4NrT_I6xgilZ8P9BkP:biyx1h3) ([raw file](https://github.com/sillsdev/liblcm/blob/f4232a846d90564fff0e42d4df9cca7fbfbcc53b/tests/SIL.LCModel.Tests/RhinoMocksToMoq/Expect.cs#L1)):*

:vomiting_face: 
We should discuss, this may be the best solution but I prefer to avoid inlining 3rd party code.
</blockquote></details>

Yeah I tried using the library but was getting errors due to the Nuget dlls not being signed, so I decided to inline it. I'd be happy to use the package if you know what we should do about that.

hahn-kev:
@jasonleenaylor when I run with the nuget package to sub RhinoMocks for Moq I get this error at runtime in dotnet framework:
```
  System.IO.FileLoadException : Could not load file or assembly 'RhinoMocksToMoq, Version=0.2.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. A strongly-named assembly is required. (Exception from HRESULT: 0x80131044)
```

in addition the default configuration of Moq doesn't match the default config of RhinoMocks (automatically setting up all properties) and the nuget replacment does not configure that properly, so we'd have to touch every place that uses MockRepository to get a mock or stub.

If you can figure out the strongly named assembly issue then I'll take the time to get the tests working by making sure setup is called before the test code uses the mock. I've pushed the version that uses the nuget package.

hahn-kev:
@jasonleenaylor I inlined the helper lib for RhinoMocks like we talked about on Friday/Thursday, I'm ready to merge this in. Not sure why that PR check is failing but I don't think it's related to this PR

jasonleenaylor (approved):
:lgtm:

> Reviewed 1 of 11 files at r1, 1 of 12 files at r2, 5 of 5 files at r4, 9 of 9 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/306#-:-O5NxjKODp79YvllW7DH:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @hahn-kev)

PR #309, opened 2024-08-16
migrate from NHunspell to WeCantSpell.Hunspell

hahn-kev:
Right now LCM is using [NHunspell](https://www.nuget.org/packages/NHunspell), which hasn't been updated in 9 years and only supports dotnet framework (it does work on windows through compatibility, but not any other platform).

It looks like the suggested alternative is [WeCantSpell.Hunspell](https://github.com/aarondandy/WeCantSpell.Hunspell).

One issue with WeCantSpell is that it makes the wordlist read only once it's been parsed. There's a discussion about that [here](https://github.com/aarondandy/WeCantSpell.Hunspell/issues/41), from what I can tell that's the biggest hurdle blocking us from adopting the library. I gave it a try, but I don't understand some of the Hunspell details enough to figure out what's going on in our old version.

Waiting for #306 to be merged in so the tests run.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/309)


aarondandy:
If it is of interest to you, I added "Add" methods in the following release: https://www.nuget.org/packages/WeCantSpell.Hunspell/6.0.0 .

jasonleenaylor (commented):
I don't see the updated WordList being serialized back out to disk for WeCantSpell, am I missing it? (I'm pretty sure that we do with NHunspell because we end up suggesting people use that dictionary in LibreOffice ect.

> Reviewed 5 of 7 files at r1, 2 of 2 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/309#-:-OKbX789EjeulK3tQ2h8:b-2cjalq)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @hahn-kev)

___
*[`src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs` line 93 at r2](https://reviewable.io/reviews/sillsdev/liblcm/309#-OKbVYyL2mwIPX2WRPat:-OKbVYyL2mwIPX2WRPau:b4m68n8) ([raw file](https://github.com/sillsdev/liblcm/blob/a0f0789eb617314c0a81145d277b0694f51bd324/src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs#L93)):*
> ```Smalltalk
> 
> 		private bool _isVernacular;
> 		private bool _gotIsVernacular;
> ```

Real question. Now that we have the option which is better, two bools or one bool?

aarondandy:
> I don't see the updated WordList being serialized back out to disk for WeCantSpell, am I missing it?

WeCantSpell doesn't support that. Add operates on the in-memory internal state only. It would be a huge pile of problems to solve to save after modification. For example, should comments be preserved on save? What about the ordering of lines in the dic/aff files? What about any simplifications or fixes that occur during load, should those be preserved? OH HECKIN' HECK what about the file encodings 🫠. Just ... so much to stress over 😬. Does NHunspell support saving? I never noticed that in their API.

To work with v6 , you would have to keep something like:
- base.aff
- base.dic
- stuff-to-add.txt

On load, load "base", then add words from your own file. When adding a word in memory, add it to both the WordList and "stuff-to-add.txt", either directly to the file or in to an in memory `List<whtvs>` representation to be persisted later.

> Real question. Now that we have the option which is better, two bools or one bool?

If you are asking me:

```csharp
private bool? _isVernacular;
public bool IsVernacular
{
  get
  {
    _isVernacular ??= Check(SpellingHelper.PrototypeWord);
    return _isVernacular.GetValueOrDefault();
  }
}

```

hahn-kev:
@jasonleenaylor I think saving changes is handled outside NHunspell in an Exception file which you can see here
https://github.com/sillsdev/liblcm/blob/aee8c1a7a9c3e3d8bed4319175e921dae6387f1f/src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs#L98-L142
`SetStatusInternal` is abstract and implemented by the specific library, this is basically exactly what aarondandy was suggesting. There's no code that I can see which is saving from NHunspell, I'm not even sure how you would do that given it's API.

hahn-kev:
I've updated this PR to remove NHunspell, and remove the strong name requirement, we're no longer strong naming LCM dlls. I'm not sure if this would qualify as a breaking change. If there's a program using LCM which requires strong named assemblies then this would be breaking, but I wouldn't expect that. As it is there's a bunch of [solutions on nuget](https://www.nuget.org/packages/strongnamer) to sign other nuget packages automatically to work around this problem, so anyone downstream that still wants strong names I would recommend that.

imnasnainaec (commented):
> Reviewed 13 of 17 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/309)* status: 15 of 19 files reviewed, all discussions resolved (waiting on @jasonleenaylor)

___
*[`tests/SIL.LCModel.Core.Tests/App.config` line 7 at r5](https://reviewable.io/reviews/sillsdev/liblcm/309#-OLF8qol8jPm94jyySjS:-OLF8qom9ZRSguiR6ZjW:bdqn22x) ([raw file](https://github.com/sillsdev/liblcm/blob/aee8c1a7a9c3e3d8bed4319175e921dae6387f1f/tests/SIL.LCModel.Core.Tests/App.config#L7)):*
> ```XML
> 			<listeners>
> 				<clear />
> 				<add name="FwTraceListener" type="SIL.LCModel.Utils.EnvVarTraceListener, SIL.LCModel.Utils, Culture=neutral, PublicKeyToken=f245775b81dcfaab" initializeData="assertuienabled='true' logfilename='%temp%/asserts.log'" />
> ```

What did `Culture=neutral` do and why isn't it needed anymore?
___
*[`src/SIL.LCModel.Core/SpellChecking/SpellEngineWeCantSpell.cs` line 0 at r5](https://reviewable.io/reviews/sillsdev/liblcm/309#-OLFC-hZEjU77X9g0bCe:-OLFC-hZEjU77X9g0bCf:b9r51gi) ([raw file](https://github.com/sillsdev/liblcm/blob/045d5b249e65d7214e3f4948a00cc165d92e5d43/src/SIL.LCModel.Core/SpellChecking/SpellEngineWeCantSpell.cs#L0)):*
`.editorconfig` has `max_line_length = 98` but this file has a number of lines longer than that.
___
*[`src/SIL.LCModel.Core/SpellChecking/SpellEngineWeCantSpell.cs` line 1 at r5](https://reviewable.io/reviews/sillsdev/liblcm/309#-OLFB1TyEn5x3j0h9wG_:-OLFB1TyEn5x3j0h9wGa:b-fcuvg3) ([raw file](https://github.com/sillsdev/liblcm/blob/045d5b249e65d7214e3f4948a00cc165d92e5d43/src/SIL.LCModel.Core/SpellChecking/SpellEngineWeCantSpell.cs#L1)):*
> ```Smalltalk
> using System;
> ```

It looks like `using System;` isn't needed.
___
*[`Directory.Build.props` line 18 at r5](https://reviewable.io/reviews/sillsdev/liblcm/309#-OLF5sXK0brWQDYGyc0s:-OLF5sXK0brWQDYGyc0t:b-3r19r5) ([raw file](https://github.com/sillsdev/liblcm/blob/045d5b249e65d7214e3f4948a00cc165d92e5d43/Directory.Build.props#L18)):*
> ```XML
>     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
>     <PackageOutputPath>$(MSBuildThisFileDirectory)/artifacts</PackageOutputPath>
> 	<WarningsAsErrors>NU1605;CS8002</WarningsAsErrors>
> ```

This file has a mix of space and tab indenting.
___
*[`src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs` line 11 at r5](https://reviewable.io/reviews/sillsdev/liblcm/309#-OLFAsrk0vCjSwerNeKo:-OLFAsrk0vCjSwerNeKp:b-bn62gl) ([raw file](https://github.com/sillsdev/liblcm/blob/045d5b249e65d7214e3f4948a00cc165d92e5d43/src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs#L11)):*
> ```Smalltalk
> using System.Text;
> using Icu;
> using SIL.PlatformUtilities;
> ```

Looks like `using SIL.PlatformUtilities;` is no longer needed.
___
*[`tests/SIL.LCModel.Core.Tests/SpellChecking/SpellingHelperTests.cs` line 8 at r5](https://reviewable.io/reviews/sillsdev/liblcm/309#-OLFBbR6-3eaMgRMyDLL:-OLFBbR6-3eaMgRMyDLM:br0czwo) ([raw file](https://github.com/sillsdev/liblcm/blob/045d5b249e65d7214e3f4948a00cc165d92e5d43/tests/SIL.LCModel.Core.Tests/SpellChecking/SpellingHelperTests.cs#L8)):*
> ```Smalltalk
> using System.IO;
> using System.Linq;
> using System.Runtime.InteropServices;
> ```

Is `using System.Runtime.InteropServices;` still needed?
___
*[`src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs` line 94 at r2](https://reviewable.io/reviews/sillsdev/liblcm/309#-OLFFnjz1LbdtnBLmq_v:-OLFFnjz1LbdtnBLmq_w:b-6dxh2y) ([raw file](https://github.com/sillsdev/liblcm/blob/a0f0789eb617314c0a81145d277b0694f51bd324/src/SIL.LCModel.Core/SpellChecking/SpellEngine.cs#L94)):*
> ```Smalltalk
> 		private bool _isVernacular;
> 		private bool _gotIsVernacular;
> 		public bool IsVernacular
> ```

According to https://stackoverflow.com/a/78230896/10210583, an empty `/// <inheritdoc /> is useless; since you've dropped it from `IsVernacular`, you might as well drop the other ones in this file.

aarondandy:
> I'm not sure if this would qualify as a breaking change.

It technically is a breaking change and is why I don't want to get myself involved in any strong naming 😅. I think you should be careful about doing this but that doesn't mean you can't obviously. If you understand your user base well enough (maybe your user base is you) it makes it a lot easier. It has been a long while since I have had to deal with GAC, strong naming, or Full Framework dependencies but I think that is where you may run into most issues. That said I haven't researched it fully myself and I forgot everything I might have known about it a decade ago... and I never looked back!

> What did Culture=neutral do and why isn't it needed anymore?

This along with the public key token is part of the assembly's fully qualified name. If you aren't doing strong naming, you don't need that.

imnasnainaec (commented):
> Reviewed 2 of 17 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/309)* status: 17 of 19 files reviewed, all discussions resolved (waiting on @jasonleenaylor)

hahn-kev:
Thanks for the review @imnasnainaec, I've applied changes from your feedback

imnasnainaec (commented):
> Reviewed 3 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/309)* status: 18 of 19 files reviewed, all discussions resolved (waiting on @jasonleenaylor)

___
*[`Directory.Build.props` line 18 at r5](https://reviewable.io/reviews/sillsdev/liblcm/309#-OLF5sXK0brWQDYGyc0s:-OLd-nq44geXUkS0HHws:b-2eh19v) ([raw file](https://github.com/sillsdev/liblcm/blob/045d5b249e65d7214e3f4948a00cc165d92e5d43/Directory.Build.props#L18)):*
<details><summary><i>Previously, imnasnainaec (D. Ror.) wrote…</i></summary><blockquote>

This file has a mix of space and tab indenting.
</blockquote></details>

It still does.

hahn-kev:
Alright @imnasnainaec it turns out the props file wasn't caught by the editor config which is why my editor was formatting it incorrectly. I've fixed that and the whitespace.

imnasnainaec (commented):
> Reviewed 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/309)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @hahn-kev)

PR #313, opened 2024-10-18
Add T Create(Guid) to ILcmFactory<T>

hahn-kev:
Right now to make a given object with a specific GUID you need to use a specialized factory method, this PR adds a low lever API to make any object with a predefined GUID.

There was a conflict with ICmPictureFactory because it declared it's own version with the same signature which I then had to remove from the actual CmPictureFactory.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/313)

### DIFF
--- a/src/SIL.LCModel/DomainImpl/FactoryAdditions.cs
+++ b/src/SIL.LCModel/DomainImpl/FactoryAdditions.cs
@@ -2137,20 +2137,6 @@ namespace SIL.LCModel.DomainImpl
 					return;
 			}
 		}
-
-		/// <summary>
-		/// Create a new entry with the given guid.
-		/// </summary>
-		public ICmPicture Create(Guid guid)
-		{
-			if (guid == Guid.Empty)
-				return Create();
-
-			int hvo = ((IDataReader) m_cache.ServiceLocator.GetInstance<IDataSetup>())
-				.GetNextRealHvo();
-			return new CmPicture(m_cache, hvo, guid);
-		}
-
 	}
 	#endregion
 
--- a/src/SIL.LCModel/FactoryInterfaceAdditions.cs
+++ b/src/SIL.LCModel/FactoryInterfaceAdditions.cs
@@ -845,11 +845,6 @@ namespace SIL.LCModel
 		ICmPicture Create(string sFolder, int anchorLoc, IPictureLocationBridge locationParser,
 			string sDescription, string srcFilename, string sLayoutPos, string sLocationRange,
 			string sCopyright, string sCaption,	PictureLocationRangeType locRangeType, string sScaleFactor);
-
-		/// <summary>
-		/// Create a new entry with the given guid.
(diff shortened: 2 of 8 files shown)


jasonleenaylor:
*[`src/SIL.LCModel/LcmGenerate/factory.vm.cs` line 67 at r1](https://reviewable.io/reviews/sillsdev/liblcm/313#-O9VVMiWEC03bGyTtIYV:-O9VVMiWEC03bGyTtIYW:b-llvj4t) ([raw file](https://github.com/sillsdev/liblcm/blob/fc2d984931fc977c71c906c5bd4c9af900aa3492/src/SIL.LCModel/LcmGenerate/factory.vm.cs#L67)):*
> ```Smalltalk
> 				throw new InvalidOperationException("Can not create more than one ${className}");
> #end
> 			if (guid == Guid.Empty) guid = Guid.NewGuid();
> ```

Is this a real use case we want to support? I can't imagine a case where you want to create an unowned object with a guid that you don't know...
I'd throw instead.

jasonleenaylor (commented):
So after looking at this and seeing that you only had to remove one I realized the more common pattern for this is the create which takes a Guid and an owner. It also you might something like that in practice. See the FactoryAdditions for LexRefTypeFactory for an example.

> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/313#-:-O9VX2Xh4fVSHEQDxHpf:b-x4emvl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @hahn-kev)

hahn-kev:
> *[`src/SIL.LCModel/LcmGenerate/factory.vm.cs` line 67 at r1](https://reviewable.io/reviews/sillsdev/liblcm/313#-O9VVMiWEC03bGyTtIYV:-O9VVMiWEC03bGyTtIYW:b-llvj4t) ([raw file](https://github.com/sillsdev/liblcm/blob/fc2d984931fc977c71c906c5bd4c9af900aa3492/src/SIL.LCModel/LcmGenerate/factory.vm.cs#L67)):*
> > ```Smalltalk
> > 				throw new InvalidOperationException("Can not create more than one ${className}");
> > #end
> > 			if (guid == Guid.Empty) guid = Guid.NewGuid();
> > ```
> 
> Is this a real use case we want to support? I can't imagine a case where you want to create an unowned object with a guid that you don't know...
> I'd throw instead.
> 
> 
> 
> 

Well I didn't support it originally, but the tests for ICmPicture had an explicit test for passing in an empty GUID. So in order to not break existing code I kept that test passing.

jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/313)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @hahn-kev)

PR #317, opened 2025-02-28
NRT Migration Phase 1

josephmyers:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/317)

### DIFF
--- a/src/CSTools/Tools/genbase.cs
+++ b/src/CSTools/Tools/genbase.cs
@@ -8,6 +8,7 @@
 using System;
 using System.IO;
 using System.Text;
+using System.Diagnostics.CodeAnalysis;
 using YYClass;
 
 namespace SIL.LCModel.Tools
@@ -169,12 +170,8 @@ namespace SIL.LCModel.Tools
 			}
 			m_outFile.WriteLine("}");
 		}
-		/// <summary>
-		///
-		/// </summary>
-		/// <param name="n"></param>
-		/// <param name="p"></param>
-		/// <param name="str"></param>
+		
+		[DoesNotReturn]
 		public void Error(int n,int p,string str)
 		{
 			Console.WriteLine(str);
@@ -185,6 +182,7 @@ namespace SIL.LCModel.Tools
 			}
 			erh.Error(new CSToolsFatalException(n,line(p),position(p),"",str));
 		}
+
 		public virtual int line(int pos) { return 0; }
 		public virtual int position(int pos) { return pos; }
 		public virtual string Saypos(int pos) { return "at "+pos+": "; }
--- a/src/CSTools/Tools/lexer.cs
+++ b/src/CSTools/Tools/lexer.cs
@@ -6,7 +6,9 @@ using System.Collections;
 using System.IO;
 using System.Text;
 using System.Globalization;
+using System.Diagnostics.CodeAnalysis;
 //using System.Runtime.Serialization.Formatters.Binary;
+
 
 namespace SIL.LCModel.Tools
 {
@@ -384,7 +386,7 @@ namespace SIL.LCModel.Tools
 				yyl.tokens.erh.Error(new CSToolsException(6,yyl,cls_name,String.Format("no factory for {0}",cls_name)));
 			try
 			{
-				return cr(yyl);
+				return cr!(yyl);
 			}
 			catch (CSToolsException x)
 			{
@@ -605,7 +607,7 @@ namespace SIL.LCModel.Tools
 				s = (SYMBOL)d;
 			try
 			{
-				rv =(int)d;
+				rv =(int)d!;
 			}
 			catch(Exception e)
 			{
@@ -638,8 +640,8 @@ namespace SIL.LCModel.Tools
 				string s = string.Format("No parsinginfo for symbol {0}",yyname());
 				syms.erh.Error(new CSToolsFatalException(9,yylx,yyname(),s));
 			}
-			bool r = pi.m_parsetable.Contains(snum);
-			entry = r?((ParserEntry)pi.m_parsetable[snum]):null;
+			bool r = pi != null && pi.m_parsetable.Contains(snum);
+			entry = pi != null && r ? ((ParserEntry)pi.m_parsetable[snum]) : null;
 			return r;
 		}
 		public virtual string yyname() { return "SYMBOL"; }
@@ -675,8 +677,8 @@ namespace SIL.LCModel.Tools
 			ParsingInfo pi = (ParsingInfo)syms.literalInfo[m_str];
 			if (pi==null)
 				syms.erh.Error(new CSToolsException(10,yylx,m_str,String.Format("Parser does not recognise literal <{0}>",m_str)));
-			bool r = pi.m_parsetable.Contains(snum);
-			entry = r?((ParserEntry)pi.m_parsetable[snum]):null;
+			bool r = pi != null && pi.m_parsetable.Contains(snum);
+			entry = pi != null && r ? ((ParserEntry)pi.m_parsetable[snum]) : null;
 			return r;
 		}
 		public override bool IsTerminal() { return true; }
@@ -970,21 +972,23 @@ namespace SIL.LCModel.Tools
 		public virtual void Error(Exception e)
 		{
 			counter++;
-			if (throwExceptions || (e is CSToolsFatalException))
-				throw(e);
-			if (e is CSToolsException)
+			if (throwExceptions)
+				throw e;
+			if (e is CSToolsException x)
 			{
-				CSToolsException x = (CSToolsException)e;
 				if (x.handled)
 					return;
 				x.handled = true;
-//				if (x.nLine !=0 || x.nChar!=0)
-//				{
-//					Console.WriteLine("line	"+x.nLine+", char "+x.nChar+": "+x.Message);
-//					return;
-//				}
+
 			}
 			Console.WriteLine(e.Message);
+		}
+
+		[DoesNotReturn]
+		public virtual void Error(CSToolsFatalException e)
+		{
+			counter++;
+			throw e;
 		}
 	}
 }
--- a/src/CSTools/Tools/lexer.cs
+++ b/src/CSTools/Tools/lexer.cs
@@ -972,7 +972,7 @@ namespace SIL.LCModel.Tools
 		public virtual void Error(Exception e)
 		{
 			counter++;
-			if (throwExceptions)
+			if (throwExceptions || (e is CSToolsFatalException))
 				throw e;
 			if (e is CSToolsException x)
 			{
--- a/src/CSTools/Tools/parser.cs
+++ b/src/CSTools/Tools/parser.cs
@@ -1554,7 +1554,9 @@ namespace SIL.LCModel.Tools
 			ParseStackEntry top = new ParseStackEntry(this,0,NextSym());
 			for (;;)
 			{
-				string cnm = top.m_value!.yyname();
+				if (top.m_value == null)
+					throw new NullReferenceException("Stack entry m_value cannot be null");
+				string cnm = top.m_value.yyname();
 				if (m_debug)
 				{
 					if (cnm.Equals("TOKEN"))
--- a/src/CSTools/Tools/parser.cs
+++ b/src/CSTools/Tools/parser.cs
@@ -1693,7 +1693,8 @@ namespace SIL.LCModel.Tools
 				yyp.m_symbols.erh.Error(new CSToolsException(16,yyp.m_lexer,"no factory for {"+cls_name+")"));
 			try
 			{
-				return cr!(yyp);
+				if (cr != null)
+					return cr(yyp);
 			}
 			catch (CSToolsException e)
 			{
--- a/src/CSTools/Tools/parser.cs
+++ b/src/CSTools/Tools/parser.cs
@@ -1504,7 +1504,7 @@ namespace SIL.LCModel.Tools
 								newtop = pe.m_action.Action(this); // before we change the stack
 							m_ungot = top.m_value;
 							Pop(ref top,((ParserReduce)pe).m_depth,er);
-							newtop.pos = top.m_value.pos;
+							newtop!.pos = top.m_value.pos;
 							top.m_value = newtop;
 						}
 						else
@@ -1554,7 +1554,7 @@ namespace SIL.LCModel.Tools
 			ParseStackEntry top = new ParseStackEntry(this,0,NextSym());
 			for (;;)
 			{
-				string cnm = top.m_value.yyname();
+				string cnm = top.m_value!.yyname();
 				if (m_debug)
 				{
 					if (cnm.Equals("TOKEN"))
@@ -1693,7 +1693,7 @@ namespace SIL.LCModel.Tools
 				yyp.m_symbols.erh.Error(new CSToolsException(16,yyp.m_lexer,"no factory for {"+cls_name+")"));
 			try
 			{
-				return cr(yyp);
+				return cr!(yyp);
 			}
 			catch (CSToolsException e)
 			{
--- a/src/CSTools/pg/pg.cs
+++ b/src/CSTools/pg/pg.cs
@@ -9,6 +9,7 @@ using System;
 using System.Collections;
 using System.IO;
 using System.Text;
+using System.Diagnostics.CodeAnalysis;
 using SIL.LCModel.Tools;
 
 public class PrecReference : TOKEN
@@ -777,12 +778,12 @@ public class ParserGenerate : SymbolsGen
 		{
 			erh.Error(new CSToolsFatalException(45,tok,string.Format("Syntax error in Parser script - possibly extra semicolon?",e.Message)));
 		}
-		m_tok = lhs;
+		m_tok = lhs!;
 		if (m_tok.IsTerminal())
 			Error(39,m_tok.pos,string.Format("Illegal left hand side <{0}> for production",m_tok.yytext));
 		if (m_symbols.m_startSymbol==null)
 			m_symbols.m_startSymbol = lhs;
-		if (lhs.m_symtype==CSymbol.SymType.unknown)
+		if (lhs!.m_symtype==CSymbol.SymType.unknown)
 			lhs.m_symtype = CSymbol.SymType.nonterminal;
 		if ((!lhs.m_defined) && lhs.m_prods.Count==0)
 		{ // lhs not defined in %symbol statement and not previously a lhs
--- a/src/CSTools/pg/pg.cs
+++ b/src/CSTools/pg/pg.cs
@@ -778,7 +778,7 @@ public class ParserGenerate : SymbolsGen
 		{
 			erh.Error(new CSToolsFatalException(45,tok,string.Format("Syntax error in Parser script - possibly extra semicolon?",e.Message)));
 		}
-		m_tok = lhs!;
+		m_tok = lhs;
 		if (m_tok.IsTerminal())
 			Error(39,m_tok.pos,string.Format("Illegal left hand side <{0}> for production",m_tok.yytext));
 		if (m_symbols.m_startSymbol==null)
--- a/src/SIL.LCModel.Core/WritingSystems/CoreWritingSystemDefinition.cs
+++ b/src/SIL.LCModel.Core/WritingSystems/CoreWritingSystemDefinition.cs
@@ -67,14 +67,14 @@ namespace SIL.LCModel.Core.WritingSystems
 			switch (e.Action)
 			{
 				case NotifyCollectionChangedAction.Add:
-					AddWordFormingOverrides(e.NewItems!.Cast<string>());
+					AddWordFormingOverrides(e.NewItems.Cast<string>());
 					break;
 				case NotifyCollectionChangedAction.Remove:
-					RemoveWordFormingOverrides(e.OldItems!.Cast<string>());
+					RemoveWordFormingOverrides(e.OldItems.Cast<string>());
 					break;
 				case NotifyCollectionChangedAction.Replace:
-					RemoveWordFormingOverrides(e.OldItems!.Cast<string>());
-					AddWordFormingOverrides(e.NewItems!.Cast<string>());
+					RemoveWordFormingOverrides(e.OldItems.Cast<string>());
+					AddWordFormingOverrides(e.NewItems.Cast<string>());
 					break;
 				case NotifyCollectionChangedAction.Reset:
 					m_wordFormingOverrides.Clear();
--- a/src/SIL.LCModel.Core/WritingSystems/CoreWritingSystemDefinition.cs
+++ b/src/SIL.LCModel.Core/WritingSystems/CoreWritingSystemDefinition.cs
@@ -67,12 +67,16 @@ namespace SIL.LCModel.Core.WritingSystems
 			switch (e.Action)
 			{
 				case NotifyCollectionChangedAction.Add:
+					if (e.NewItems == null) throw new NullReferenceException("e.NewItems");
 					AddWordFormingOverrides(e.NewItems.Cast<string>());
 					break;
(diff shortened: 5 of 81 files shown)


src/CSTools/Tools/lexer.cs
```diff
@@ -970,21 +972,23 @@ public ErrorHandler() {}
 		public virtual void Error(Exception e)
 		{
 			counter++;
-			if (throwExceptions || (e is CSToolsFatalException))
```
hahn-kev, line 973: we probably still need this as `Error` could be called with a `CSToolsFatalException` but if the caller doesn't know the type then the new method `Error(CSToolsFatalException e)` won't be called.

src/CSTools/Tools/parser.cs
```diff
@@ -1504,7 +1504,7 @@ protected bool Error(ref ParseStackEntry top, string str)
 								newtop = pe.m_action.Action(this); // before we change the stack
 							m_ungot = top.m_value;
 							Pop(ref top,((ParserReduce)pe).m_depth,er);
-							newtop.pos = top.m_value.pos;
+							newtop!.pos = top.m_value.pos;
```
hahn-kev, line 1507: wont this throw if `pe.m_action` is null?

src/CSTools/Tools/parser.cs
```diff
@@ -1554,7 +1554,7 @@ SYMBOL Parse()
 			ParseStackEntry top = new ParseStackEntry(this,0,NextSym());
 			for (;;)
 			{
-				string cnm = top.m_value.yyname();
+				string cnm = top.m_value!.yyname();
```
hahn-kev, line 1557: this could be null because `top` is reassigned in the call to `Pop(ref top ...` meaning `m_value` could be null here. I say we just explicitly throw a null error before this line if `top.m_value` is null as that's what it would have done before.

src/CSTools/Tools/parser.cs
```diff
@@ -1693,7 +1693,7 @@ public static object create(string cls_name,Parser yyp)
 				yyp.m_symbols.erh.Error(new CSToolsException(16,yyp.m_lexer,"no factory for {"+cls_name+")"));
 			try
 			{
-				return cr(yyp);
+				return cr!(yyp);
```
hahn-kev, line 1696: here's the same as a previous comment, we should just skip this if it's null as the previous check may not throw

src/CSTools/pg/pg.cs
```diff
@@ -777,12 +778,12 @@ internal void ParseProduction()
 		{
 			erh.Error(new CSToolsFatalException(45,tok,string.Format("Syntax error in Parser script - possibly extra semicolon?",e.Message)));
 		}
-		m_tok = lhs;
+		m_tok = lhs!;
```
hahn-kev, line 781: this is no longer needed since erh.Error with a Fatal Exception throws.

src/CSTools/Tools/genbase.cs
```diff
@@ -169,12 +170,8 @@ public void EmitClassDefin(string b,ref int p,int max,CsReader inf,string defbas
...
 			m_outFile.WriteLine("}");
 		}
-		/// <summary>
-		///
-		/// </summary>
-		/// <param name="n"></param>
-		/// <param name="p"></param>
-		/// <param name="str"></param>
+		
+		[DoesNotReturn]
```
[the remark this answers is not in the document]
josephmyers: It does, yes, when you give it the fatal exception

src/CSTools/Tools/lexer.cs
```diff
@@ -384,7 +386,7 @@ public static object create(string cls_name,Lexer yyl)
 				yyl.tokens.erh.Error(new CSToolsException(6,yyl,cls_name,String.Format("no factory for {0}",cls_name)));
 			try
 			{
-				return cr(yyl);
+				return cr!(yyl);
```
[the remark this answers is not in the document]
josephmyers: There are two reasons I didn't go that route:
1) The potential for an exception is handled just below, seemingly intentionally
2) This is a tool, so I'm assuming that the above is acceptable

We can go your route, as well.

src/SIL.LCModel.Core/WritingSystems/CoreWritingSystemDefinition.cs
```diff
@@ -67,14 +67,14 @@ private void MainCharacterSetChanged(object sender, NotifyCollectionChangedEvent
 			switch (e.Action)
 			{
 				case NotifyCollectionChangedAction.Add:
-					AddWordFormingOverrides(e.NewItems.Cast<string>());
+					AddWordFormingOverrides(e.NewItems!.Cast<string>());
```
hahn-kev, line 70: these can be null so we should just throw manually if it is

src/SIL.LCModel/Application/ApplicationServices/XmlImportData.cs
```diff
@@ -593,7 +590,7 @@ private void CopyCustomFieldData(ICmObject cmoOld, ICmObject cmoNew)
 			Debug.Assert(mdc != null);
 			ISilDataAccessManaged sda = m_cache.DomainDataByFlid as ISilDataAccessManaged;
 			Debug.Assert(sda != null);
-			foreach (int flid in mdc.GetFields(cmoOld.ClassID, true, (int)CellarPropertyTypeFilter.All))
+			foreach (int flid in mdc!.GetFields(cmoOld.ClassID, true, (int)CellarPropertyTypeFilter.All))
```
hahn-kev, line 593: there's a debug assert a few lines up, let's just replace that with an exception since one would be thrown here anyway if it was null

src/SIL.LCModel/Application/ApplicationServices/XmlImportData.cs
```diff
@@ -1005,7 +1002,7 @@ private void ReadXmlObject(XmlReader xrdr, FieldInfo fi, ICmObject objToUse)
 					else
 						cmo = factory.Create(guid, (IPhPhonemeSet)fi.Owner);
 					// Remove the default code added by PhTerminalUnit.SetDefaultValuesAfterInit in OverridesLing_Lex.
-					(cmo as PhBdryMarker).CodesOS.Clear();
+					((PhBdryMarker)cmo).CodesOS.Clear();
```
hahn-kev, line 1005: there's another place in this method where it does a similar pattern, but instead of casting it does this:
```suggestion
					(cmo as PhBdryMarker)?.CodesOS.Clear();
```

src/CSTools/Tools/lexer.cs
josephmyers: You're right that there's a problem, but I was thrown off when you mentioned the caller not knowing the type. Whether they know the type or not doesn't matter (I don't think), what matters is that C# resolves its methods at compile-time. So the problem with this code is that 

				CSToolsException e = new CSToolsFatalException(...);
				syms.erh.Error(e);

calls the less derived method. Or maybe there's another problem?

src/CSTools/Tools/parser.cs
josephmyers: Yes, it will. The way I interpret this section is, "if pe is non-null, so is pe.m_action"

src/CSTools/Tools/parser.cs
josephmyers: But it returns after that call. It shouldn't ever do another iteration in this loop if it gets the successful parse. Right?

src/SIL.LCModel.Core/WritingSystems/CoreWritingSystemDefinition.cs
josephmyers: I'm not sure why I added these. When I remove them, there are no warnings. I get that they can be null, based on the spec, but I'm not sure they could ever cause a null exception in their respective case statements.

src/SIL.LCModel/DomainImpl/OverridesLing_Lex.cs
```diff
@@ -5838,13 +5837,13 @@ public string SenseNumber
 
 				if (Owner is ILexEntry)
 				{
-					var le = Owner as ILexEntry;
+					var le = (ILexEntry)Owner;
```
hahn-kev, line 5840: I'd shift this into the `if`

src/SIL.LCModel/DomainImpl/OverridesLing_MoClasses.cs
```diff
@@ -504,14 +504,12 @@ public void SwitchReferences(IMoMorphSynAnalysis sourceMsa)
...
-					IWfiMorphBundle wmb = obj as IWfiMorphBundle;
-					Debug.Assert(wmb.MsaRA == sourceMsa);
+					var wmb = (IWfiMorphBundle)obj;
 					wmb.MsaRA = this;
 				}
 				else if (obj is ILexSense)
 				{
-					ILexSense sense = obj as ILexSense;
-					Debug.Assert(sense.MorphoSyntaxAnalysisRA == sourceMsa);
+					var sense = (ILexSense)obj;
```
hahn-kev, line 512: can be shifted into the if

src/SIL.LCModel/DomainImpl/OverridesLing_MoClasses.cs
```diff
@@ -504,14 +504,12 @@ public void SwitchReferences(IMoMorphSynAnalysis sourceMsa)
 			{
 				if (obj is IWfiMorphBundle)
 				{
-					IWfiMorphBundle wmb = obj as IWfiMorphBundle;
-					Debug.Assert(wmb.MsaRA == sourceMsa);
+					var wmb = (IWfiMorphBundle)obj;
```
hahn-kev, line 507: shift this into the if

src/SIL.LCModel/DomainImpl/OverridesLing_MoClasses.cs
```diff
@@ -1887,7 +1885,7 @@ public override ITsString ShortNameTSS
 			{
 				var entry = Owner as ILexEntry;
 				var tisb = TsStringUtils.MakeIncStrBldr();
-				(entry as LexEntry).CitationFormWithAffixTypeTss(tisb);
+				(entry as LexEntry)!.CitationFormWithAffixTypeTss(tisb);
```
hahn-kev, line 1888: rather than this let's just cast `(LexEntry)Owner` above

src/SIL.LCModel/DomainImpl/OverridesLing_MoClasses.cs
```diff
@@ -3434,7 +3430,7 @@ public override ITsString ShortNameTSS
 			get
 			{
 				var tss = Form.VernacularDefaultWritingSystem;
-				if (tss != null || tss.Length > 0)
+				if (tss.Length > 0)
```
hahn-kev, line 3433: I actually think the previous behavior was correct and that string can totally be null, you could just do `tss?.Lengh > 0` though

hahn-kev (commented):
left some more comments

src/SIL.LCModel/Application/ApplicationServices/XmlImportData.cs
```diff
@@ -1139,14 +1136,14 @@ private void ReadXmlObject(XmlReader xrdr, FieldInfo fi, ICmObject objToUse)
...
-				ILexEntry le = cmo as ILexEntry;
+				ILexEntry le = (ILexEntry)cmo;
 				ILexEntry leDup;
 				StoreEntryInMap(le, fNewObject, true, out leDup);
 				if (leDup != null)
 				{
 					MergeRedundantEntries(leDup, le);
 					List<WsString> delKeys = new List<WsString>();
-					foreach (var x in m_mapFormEntry)
+					foreach (var x in m_mapFormEntry!)
```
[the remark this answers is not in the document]
josephmyers: Oddly enough, this required me to upgrade to C# 9. I would've thought it wouldn't matter if we're using the Nullable package?

src/SIL.LCModel/DomainImpl/AnalysisAdjuster.cs
```diff
@@ -1511,26 +1511,29 @@ private List<IAnalysisReference> MoveReferencesFromOldPara(ISegment newSeg, int
 		{
 			Debug.Assert(m_segPartlyMoved != null, "Nothing to do! I shouldn't have been called.");
 			var refsMoved = new List<IAnalysisReference>();
-			var oldPara = (IStTxtPara)m_segPartlyMoved.Owner;
-			var oldReferences = oldPara.GetTags().Cast<IAnalysisReference>().ToList();
-			oldReferences.AddRange(oldPara.GetChartCellRefs().Cast<IAnalysisReference>());
-			foreach (var iar in oldReferences)
+			if (m_segPartlyMoved != null)
```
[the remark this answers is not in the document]
josephmyers: Yeah, good call

src/SIL.LCModel/DomainImpl/CmObject.cs
```diff
@@ -807,7 +807,7 @@ public void MergeSelectedPropertiesOfObject(ICmObject objSrc, bool fLoseNoString
 							}
 					}
 				}
-				Object myCurrentValue = null;
+				Object myCurrentValue = new object();
```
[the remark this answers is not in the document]
josephmyers: Operating under the assumption that the code is essentially working as is, all this does is fix a bunch of compiler warnings.

src/SIL.LCModel/DomainImpl/CmObject.cs
```diff
@@ -1021,9 +1020,8 @@ public void MergeSelectedPropertiesOfObject(ICmObject objSrc, bool fLoseNoString
 							var currentObj = myCurrentValue as ICmObject;
 							if (myCurrentValue == null)
 							{
-								if (nType == (int)CellarPropertyType.OwningAtomic || mySetMethod != null)
+								if ((nType == (int)CellarPropertyType.OwningAtomic && mySetMethod != null) || mySetMethod != null)
```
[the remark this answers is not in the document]
josephmyers: This one for sure made me scratch my head. If we do what you suggest, though, don't we change the behavior when `nType != (int)CellarPropertyType.OwningAtomic && mySetMethod != null`?

src/SIL.LCModel/DomainImpl/OverridesLing_Lex.cs
```diff
@@ -2472,7 +2472,7 @@ public string LIFTid
 					sLiftId = ExtractAttributeFromLiftResidue(sResidue, "id");
 				if (String.IsNullOrEmpty(sLiftId))
 					return HeadWord.Text + "_" + Guid;
-				int idx = sLiftId.IndexOf('_');
+				int idx = sLiftId!.IndexOf('_');
```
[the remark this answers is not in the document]
josephmyers: It is according to the compiler

src/SIL.LCModel/DomainImpl/OverridesLing_Lex.cs
```diff
@@ -3428,7 +3428,7 @@ public override void MergeObject(ICmObject objSrc, bool fLoseNoStringData)
 			var homoForm = HomographFormKey;
 			var le = objSrc as ILexEntry;
 			// If the lexeme forms don't match, and they both have content in the vernacular, make the other LF an allomorph.
-			if (LexemeFormOA != null && le.LexemeFormOA != null &&
+			if (LexemeFormOA != null && le!.LexemeFormOA != null &&
```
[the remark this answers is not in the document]
josephmyers: I think I got it. This file is so massive Rider is saying it can't analyze it.

src/SIL.LCModel/DomainImpl/OverridesLing_Lex.cs
```diff
@@ -5239,7 +5238,7 @@ public string LIFTid
 					sLiftId = LexEntry.ExtractAttributeFromLiftResidue(sResidue, "id");
 				if (String.IsNullOrEmpty(sLiftId))
 					return Guid.ToString();
-				int idx = sLiftId.IndexOf('_');
+				int idx = sLiftId!.IndexOf('_');
```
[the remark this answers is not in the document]
josephmyers: Yep

src/SIL.LCModel/DomainImpl/OverridesLing_MoClasses.cs
josephmyers: Thanks. It's weird that it didn't flag this. That's probably why I removed it

src/SIL.LCModel.Core/WritingSystems/CoreWritingSystemDefinition.cs
josephmyers: Ok, `dotnet build` produces a warning, but the IDE doesn't. That's why they were there. I'll add the explicit checks

src/CSTools/Tools/lexer.cs
hahn-kev: Yes that's what I was referring to. In this case the 2nd line doesn't know the actual type, just that it's a tools exception.

src/CSTools/Tools/parser.cs
hahn-kev: oh you're right, man this code is just confusing. I'd still be inclined to throw as that's always going to be a better error then what gets thrown automatically, and then it's obvious to anyone looking at the code what will happen if it's null here.

src/SIL.LCModel/Application/ApplicationServices/XmlImportData.cs
```diff
@@ -605,7 +603,7 @@ private void CopyCustomFieldData(ICmObject cmoOld, ICmObject cmoNew)
 					case CellarPropertyType.Boolean:
 						break;
 					case CellarPropertyType.Integer:
-						int val = sda.get_IntProp(cmoOld.Hvo, flid);
+						int val = sda!.get_IntProp(cmoOld.Hvo, flid);
```
hahn-kev, line 606: now that you're throwing when sda is null I don't think you need `!` here or any of the other lines below

src/SIL.LCModel/DomainImpl/CmObject.cs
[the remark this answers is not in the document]
hahn-kev: true, so maybe we do this `if (nType == (int)CellarPropertyType.OwningAtomic && mySetMethod == null) throw error`. before this if. Then we just simplify this to `if (mySetMethod != null)`. 
The reason being that I read this code as saying OwningAtomic requires a mySetMethod (seeing as how there's an assert debug followed by using the method. So the original behavior would be preserved we would just be more explicit about throwing the error when null.

src/SIL.LCModel/DomainImpl/OverridesLing_Wfi.cs
```diff
@@ -612,7 +612,7 @@ protected override void OnBeforeObjectDeleted()
 		{
 			var repo = Services.GetInstance<IWfiWordformRepository>() as IWfiWordformRepositoryInternal;
 			foreach (int ws in Form.AvailableWritingSystemIds)
-				repo.RemoveForm(Form.get_String(ws), ws);
+				repo?.RemoveForm(Form.get_String(ws), ws);
```
hahn-kev, line 615: I say we just cast `repo` above rather than not calling remove if it's the wrong type

src/SIL.LCModel/DomainImpl/ScrTxtPara.cs
```diff
@@ -1945,7 +1945,7 @@ protected override void OnContentsChanged(ITsString originalValue, ITsString new
 
 			if ((originalValue == null && newValue != null) ||
 				(originalValue != null && newValue == null) ||
-				(originalValue != null && originalValue.Text != newValue.Text))
+				(originalValue != null && originalValue.Text != newValue!.Text))
```
hahn-kev, line 1948: so we only do this when something is not null, or if they're not equal. Could we do this `if (!(originalValue == null && newValue == null) && originalValue?.Text != newValue?.Text)`? is that any better? if both are null then we don't succeed. And then one of the values will be defined so we can compare to a maybe null one.

src/SIL.LCModel/DomainServices/DataMigration/DataMigration7000010.cs
```diff
@@ -2171,7 +2171,7 @@ private static XElement AddSegmentAnalyses(
 			if (xficsForCurrentOldSegment.Count > 0)
 			{
 				foreach (var key in xficsForCurrentOldSegment.Keys)
-					xficsForCurrentPara.Remove(key);
+					xficsForCurrentPara?.Remove(key);
```
hahn-kev, line 2174: this is a functional change, if `xficsForCurrentPara` is null then nothing will happen, previously it would throw.

tests/SIL.LCModel.Core.Tests/SIL.LCModel.Core.Tests.csproj
```diff
@@ -6,6 +6,7 @@
     <Description>The liblcm library is the core FieldWorks model for linguistic analyses of languages. Tools in this library provide the ability to store and interact with language and culture data, including anthropological, text corpus, and linguistics data.
 This package provides unit tests for SIL.LCModel.Core.</Description>
     <IsPackable>true</IsPackable>
+	  <Nullable>disable</Nullable>
```
hahn-kev, line 9: indenting is weird

tests/SIL.LCModel.FixData.Tests/SIL.LCModel.FixData.Tests.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFrameworks>net461;net8.0</TargetFrameworks>
     <RootNamespace>SIL.LCModel.FixData</RootNamespace>
     <IsPackable>false</IsPackable>
+	  <Nullable>disable</Nullable>
```
hahn-kev, line 7: indenting is weird

hahn-kev (approved):
looks good to me, I think there's still a couple comments to resolve. I left 2 comments about weird indenting in the test projects but they're all weird.

src/CSTools/Tools/lexer.cs
josephmyers: Right, I get what you're saying now!

src/CSTools/Tools/parser.cs
josephmyers: Sounds good

src/SIL.LCModel/DomainImpl/CmObject.cs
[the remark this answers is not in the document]
josephmyers: Sounds good. Done

src/SIL.LCModel/Application/ApplicationServices/XmlImportData.cs
josephmyers: Done

src/SIL.LCModel/DomainImpl/OverridesLing_Wfi.cs
josephmyers: Done

src/SIL.LCModel/DomainImpl/ScrTxtPara.cs
josephmyers: Sure

src/SIL.LCModel/DomainServices/DataMigration/DataMigration7000010.cs
josephmyers: Good point, I thought it was acceptable to maintain the ability to continue, but your way is more accurate

tests/SIL.LCModel.Core.Tests/SIL.LCModel.Core.Tests.csproj
josephmyers: Ah, I must have some Rider settings weird

josephmyers:
@hahn-kev I think I got everything. Thanks for approving

PR #319, opened 2025-02-28
LT-21205: Features should obey the order set in Feature Types 

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-21205 by changing LongName to sort features based on the order in the type with any remaining features sorted alphabetically afterwards.  I also made LongNameSorted use the same code as LongName.

### DIFF
--- a/src/SIL.LCModel/DomainImpl/OverridesCellar.cs
+++ b/src/SIL.LCModel/DomainImpl/OverridesCellar.cs
@@ -2460,7 +2460,7 @@ namespace SIL.LCModel.DomainImpl
 		{
 			get
 			{
-				return GetFeatureValueStringSorted();
+				return GetFeatureValueString(true);
 			}
 		}
 		/// <summary>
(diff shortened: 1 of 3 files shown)


src/SIL.LCModel/DomainImpl/OverridesCellar.cs
```diff
@@ -3994,56 +3986,21 @@ private ITsString GetFeatureValueString(bool fLongForm)
...
 		}
 
-		internal ITsString GetFeatureValueStringSorted()
+		private string FeatureSpecKey(IFsFeatureSpecification spec)
 		{
-			var tisb = TsStringUtils.MakeIncStrBldr();
-			var iCount = FeatureSpecsOC.Count;
-			if (iCount > 0)
+			if (spec.FeatureRA == null)
+				return "";
```
hahn-kev, line 3992: I'm not sure when exactly this would get hit, but it will mean this spec will be sorted to the front, and if there's more than one spec with a null RA then they will not be consistently ordered. I'd either have a fallback ordering or return something like `"2"+spec.Guid` to ensure that they're always ordered consistently.

PR #322, opened 2025-03-13
Fix LT-21729: Crash deleting phoneme features set

jtmaxwell3:
Deleting a phonological feature set from a rule caused a crash if one of the feature values is "agree" because in the process of deleting the feature set context RemoveObjectSideEffectsInternal deletes the feature values with "agree" and "disagree" from LanguageProject.PhonologicalDataOA.FeatConstraintsOS, and the code for doing that deletes the referring feature set context.  So the feature set context gets deleted twice, causing a crash.  I avoid this by removing the feature constraint from the context before deleting it from FeatConstraintsOS.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/322)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,3 @@
-
 # LCModel Library
 
 ## Description
--- a/README.md
+++ b/README.md
@@ -1,3 +1,4 @@
+
 # LCModel Library
 
 ## Description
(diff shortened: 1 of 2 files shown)


jasonleenaylor (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/322#-:-OLGBRqNB6Zg2mKpHrzk:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

PR #323, opened 2025-04-08
Fix LT-22032: Make valid character dialog more valid

jasonleenaylor:
* There was a valid class of combining character categories related to extended whitespace not yet handled

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/323)

### DIFF
--- a/src/SIL.LCModel.Core/Text/UcdProperty.cs
+++ b/src/SIL.LCModel.Core/Text/UcdProperty.cs
@@ -208,7 +208,7 @@ namespace SIL.LCModel.Core.Text
 		#region static accessor functions
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
-		/// Fills a two-tiered hash table system with all the instances of this class that
+		/// Fills a two-tiered hash table system with all the categories and classes we expect
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
 		private static void InitializeHashTables()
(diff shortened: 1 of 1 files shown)


mark-sil (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/323)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jasonleenaylor)

hahn-kev (commented):
what do all these magic numbers mean?

jasonleenaylor (commented):
The magic comes out of unicode normalization ordering that I have not bothered to dig into. While debugging I identified that calling the unicode function for these whitespace characters was returning a 2 which was not supported by the existing code. I couldn't find any resources that explained what '2' should mean, or even that it existed as a valid combining class. I eventually decided that if unorm2_getCombiningClass() could return 2 for these characters the initialization of the mapping ought to support it. We don't actually care about those values in our current uses of the code. An alternative solution would have been to use try/catch or TryGetValue for all dictionary access in UcdProperty but then I would have to pick a reasonable default behavior for each detail.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/323#-:-ONojmS63Zgr8oY4F7cu:b-hk4gd7)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jasonleenaylor)

Issue #310, opened 2024-08-17
Enable Nullable reference types


hahn-kev:
Since starting to use LCM with Fieldworks lite I'm running into places where NRT would help avoid runtime bugs. 

Read about NRT [here](https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references).

We need to plan how to roll this out, one huge PR is going to make merging difficult so it should be done in stages. Maybe we do it per project or maybe at a more fine grained level. 

Generated code will probably be the biggest area that will be tricky to figure out. 

Finally I'm not sure how this works with dotnet framework, some things I read said it should work but it requires testing. 

Finally we should consider if we want to make null issues errors or warnings. For Lexbox we found that by making them errors we actually paid attention and we largely ignored it when it was just a warning.

imnasnainaec:
(From a Claude Sonnet 4.6 planning session.)

# NRT Migration Plan

## Current State

**Completed** (merged via PR #317, March 2025):

- `Directory.Build.props` sets `<Nullable>warnings</Nullable>` globally — all projects get NRT in warning mode
- ~76 hand-written files annotated across `SIL.LCModel`, `.Core`, `.Utils`, `.Build.Tasks`, `CSTools`

**Still open:**

- NRT violations are *warnings only* — the issue author noted that errors are what actually drive adoption (Lexbox experience)
- Velocity templates are untouched — all generated code (`GeneratedInterfaces.cs`, `GeneratedClasses.cs`, etc.) is unannotated, which matters because the generated public interfaces (`ICmObject`, `ICmProject`, etc.) are exactly the API that motivated this issue
- PR #324 ("NRT Migration Phase 2") attempted project-level `enable` and more annotations but stalled due to scope creep and inconsistent AI-generated diffs; team consensus was that only the Build.Tasks, Utils, and public-interface changes are worth landing

## The Key Decision Fork

Every plan must answer two independent questions:

1. **What to do with generated code** — suppress warnings in templates, or actually annotate the templates
2. **When to escalate from `warnings` to `errors`** — and what that requires first

---

## Plan A — Suppress generated code, then escalate to errors

<details><summary>Steps:</summary>

1. In each output template (e.g. `main.vm.cs`), emit `#nullable disable` at the top of each generated `.cs` file. Silences NRT noise from generated code without touching template logic.
[...]

PR #324, opened 2025-04-17
NRT Migration Phase 2

josephmyers:
(Part of  #310)

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/324)

### DIFF
--- a/src/SIL.LCModel.Core/Phonology/PhonEnvRecognizer.cs
+++ b/src/SIL.LCModel.Core/Phonology/PhonEnvRecognizer.cs
@@ -2,6 +2,8 @@
 // This software is licensed under the LGPL, version 2.1 or later
 // (http://www.gnu.org/licenses/lgpl-2.1.html)
 
+#nullable enable
+
 using System;
 using System.Diagnostics;
 using System.Text;
@@ -16,11 +18,11 @@ namespace SIL.LCModel.Core.Phonology
 	/// </summary>
 	public class PhonEnvRecognizer
 	{
-		private bool m_fSuccess;
-		private string m_sErrorMessage;
-		private PhonEnvParser m_parser;
-		private string[] m_saNaturalClasses;
-		private string[] m_saSegments;
+		private bool m_fSuccess = false;
+		private string m_sErrorMessage = "";
+		private PhonEnvParser? m_parser;
+		private string[] m_saNaturalClasses = {};
+		private string[] m_saSegments = {};
 
 
 		/// <summary>
@@ -47,7 +49,7 @@ namespace SIL.LCModel.Core.Phonology
 		/// <param name="saSegments">The new set of segments.</param>
 		public void ResetSegments(string[] saSegments)
 		{
-			m_parser.ResetSegments(saSegments);
+			m_parser!.ResetSegments(saSegments);
 
 		}
 
@@ -57,7 +59,7 @@ namespace SIL.LCModel.Core.Phonology
 		/// <param name="saNaturalClasses"></param>
 		public void ResetNaturalClasses(string[] saNaturalClasses)
 		{
-			m_parser.ResetNaturalClasses(saNaturalClasses);
+			m_parser!.ResetNaturalClasses(saNaturalClasses);
 
 		}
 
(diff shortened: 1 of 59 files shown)


src/SIL.LCModel/Infrastructure/Impl/RepositoryAdditions.cs
```diff
@@ -103,7 +105,7 @@ public IAnalysis GetObject(int hvo)
 			return (IAnalysis)m_everythingRepos.GetObject(hvo);
 		}
 
-		public bool TryGetObject(int hvo, out IAnalysis obj)
+		public bool TryGetObject(int hvo, out IAnalysis? obj)
```
hahn-kev, line 108: I suggest for these TryGet methods we put `[MaybeNullWhen(false)]` on the out variable, then we can remove the `?`, that says when false is returned obj can be null.

src/SIL.LCModel.Core/Phonology/phonenv.parser.cs
```diff
@@ -2,6 +2,8 @@
 // This software is licensed under the LGPL, version 2.1 or later
 // (http://www.gnu.org/licenses/lgpl-2.1.html)
 
+#nullable enable
```
hahn-kev, line 5: yeah maybe let's not enable it here since we just had to put `!` all over the place, I'm not sure it's adding anything

src/SIL.LCModel.Core/Phonology/PhonEnvRecognizer.cs
```diff
@@ -57,7 +59,7 @@ public void ResetSegments(string[] saSegments)
 		/// <param name="saNaturalClasses"></param>
 		public void ResetNaturalClasses(string[] saNaturalClasses)
 		{
-			m_parser.ResetNaturalClasses(saNaturalClasses);
+			m_parser!.ResetNaturalClasses(saNaturalClasses);
```
hahn-kev, line 62: I say here and the previous method we throw an exception indicating that the class wasn't initlized properly if `m_parser` is null, otherwise a useless null exception will be thrown instead.

src/SIL.LCModel.Core/Phonology/PhonEnvRecognizer.cs
```diff
@@ -16,11 +18,11 @@ namespace SIL.LCModel.Core.Phonology
...
 	public class PhonEnvRecognizer
 	{
-		private bool m_fSuccess;
-		private string m_sErrorMessage;
-		private PhonEnvParser m_parser;
-		private string[] m_saNaturalClasses;
-		private string[] m_saSegments;
+		private bool m_fSuccess = false;
+		private string m_sErrorMessage = "";
+		private PhonEnvParser? m_parser;
```
hahn-kev, line 23: I say we just new up the parser here. It seems like it will be created when it's needed, but there's no arguments required so we should just make it here so it's never null.

josephmyers:
@hahn-kev, I know it's been a while since we last worked on this. The last thing we discussed was wanting to push this WIP so that you/we could determine how we want to proceed. This work is really all over the place, because I was trying to get various AI's to do what I wanted, finding minimal success with any. I may give this more effort today.

I remember two major directions we were considering, though there are surely more:
- Enabling nullable for only the classes directly relevant to you and your team's needs
- Enabling nullable by project, starting with the "least dependent"

I don't think fixing everything at once, which is where this PR currently is, is proving to be the most practical.

josephmyers:
I spent some time working with an updated version of Copilot today, feeding it some highly explicit instructions and focusing its scope. I was able to reach something pretty acceptable with the Build.Tasks project. I'm going to continue to apply this procedure to other projects, as long as it continues to work well

josephmyers:
So, at this point, we have a couple viable options:
- Keep everything relevant to you and your team, as well as the changes in Build.Tasks and Utils, undoing everything else
- Press on with the remaining projects (after I've finished my work for LexBox)

josephmyers:
@hahn-kev is this guy dead in the water? Last I checked, we needed something different than just a regular code review here. More of a discussion as to how we want to press forward. It may be easiest, given my current availability, to try to identify the portions of this that are acceptable and undo the rest.

hahn-kev:
Yeah I think you're right. I think the changes made to public interfaces are where the real benefit is going to be found. Everything else will benefit internal LCM work, but will take to much time to be worth it for the time being.

hahn-kev:
I think this is too big to merge in

PR #325, opened 2025-05-02
Fix LT-22110: Intermittent crash in Segment.get_AnalysesRS

jtmaxwell3:
If a Segment is no longer valid, then Segment.AnalysesRS crashes.  This tends to happen if the user edits the baseline text after analyzing some of the text.  This happens to Mark Bean a lot.  I fixed all of the places that he got a crash, but there are hundreds of references to Segent.AnalysesRS in liblcm.  It would be nice if Segment.AnalysesRS could return a dummy AnalysesRS or something when the Segment is not valid to avoid crashes.

### DIFF
--- a/src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs
+++ b/src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs
@@ -402,6 +402,7 @@ namespace SIL.LCModel.DomainServices
 			var segs = new HashSet<ISegment>();
 			foreach (ISegment seg in wordform.OccurrencesInTexts)
 			{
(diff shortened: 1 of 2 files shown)


jasonleenaylor (approved):
:lgtm:

(sigh)
More evidence that a model change which versions the text so that analyses are never invalidated is warrented.

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/325#-:-OPGwXNp9uMaizNiYJzO:b-vpi26m)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

PR #326, opened 2025-05-08
Fix LT-22121: Analysis guesser should not guess ras for rAs

jtmaxwell3:
A06-Kurdish uses A to represent a separate character from a.  But the icu locale thinks that A is the capital of a.  This means that the analysis guesser guesses analyses based on ras for rAs.  One easy way to fix this would be to change the A to U+0391 GREEK CAPITAL LETTER ALPHA, but the data is from outside the project.  An alternative is to only try lowercase analyses if the first letter of the wordform is uppercase.  This will still guess ras for RAs, but getting it to guess rAs instead would require a lot more work since there are several places in FieldWorks where it is assumed that you use ToLower to lowercase words.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/326)

### DIFF
--- a/src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs
+++ b/src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs
@@ -729,18 +729,9 @@ namespace SIL.LCModel.DomainServices
 		private IWfiWordform GetLowercaseWordform(AnalysisOccurrence occurrence, int ws,
 			bool onlyIndexZeroLowercaseMatching, IWfiWordform wordform)
 		{
-			// TODO: make it look for the first word in the sentence...may not be at Index 0!
-			if (occurrence == null)
-				return null;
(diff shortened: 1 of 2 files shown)


imnasnainaec:
Would it be worth adding a test to https://github.com/sillsdev/liblcm/blob/LT-22121/tests/SIL.LCModel.Tests/DomainServices/AnalysisGuessServicesTests.cs?

jasonleenaylor (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/326#-:-OPqE23k5vsBIRiHJntY:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

PR #327, opened 2025-06-06
Fix LT-22121: Analyze guesser should not guess ras for rAs

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-22121.  In order to handle cases where the occurrence was rAs but the wordform was ras, I changed the code to look up the wordform, the occurrence form, and the occurrence form with the first letter lowercased.  These are usually all the same, but in rare cases they can all be different if the user chooses a wordform that isn't the occurrence form or an initial lowercase form of the occurrence form.  This needed a lot of changes to the code.  This should probably be saved for release/9.3.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/327)

### DIFF
--- a/src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs
+++ b/src/SIL.LCModel/DomainServices/AnalysisGuessServices.cs
@@ -67,6 +67,17 @@ namespace SIL.LCModel.DomainServices
 			public bool lowercased = false; // whether the word form of the analysis was lowercased
 			public int priority = 0; // the priority of the count
 			public int count = 0;
+
+			public PriorityCount Clone()
+			{
+				PriorityCount priorityCount = new()
+				{
+					lowercased = lowercased,
+					priority = priority,
+					count = count
+				};
+				return priorityCount;
+			}
 		}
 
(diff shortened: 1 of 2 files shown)


jasonleenaylor:
*[`tests/SIL.LCModel.Tests/DomainServices/AnalysisGuessServicesTests.cs` line 962 at r1](https://reviewable.io/reviews/sillsdev/liblcm/327#-OStFF3s-S9jyOQ6KSeJ:-OStFF3s-S9jyOQ6KSeK:b-900dtk) ([raw file](https://github.com/sillsdev/liblcm/blob/73cd3ec74d5c29fec0ca4fbf6a14469e98f7675d/tests/SIL.LCModel.Tests/DomainServices/AnalysisGuessServicesTests.cs#L962)):*
> ```Smalltalk
> 
> 				// Don't guess lowercase even if it exists.
> 				IWfiAnalysis ras = WordAnalysisOrGlossServices.CreateNewAnalysisWAG(setup.Words_para0[21]);
> ```

For clarity I think you should move this below the first assert.

Issue #328, opened 2025-06-16
Trying to save data with a \u0002 crashes LCM


hahn-kev:
I've got a simple reproduction in my own code. Basically it looks like this:
```C#
UndoableUnitOfWorkHelper.Do("create entry", "undo", api.Cache.ServiceLocator.ActionHandler,
    () =>
    {
        var entry = api.Cache.ServiceLocator.GetInstance<ILexEntryFactory>()
            .Create(Guid.NewGuid(),
            api.Cache.ServiceLocator
                .GetInstance<ILangProjectRepository>().Singleton.LexDbOA);
        entry.LexemeFormOA = api.Cache.CreateLexemeForm();
        entry.LexemeFormOA.Form.set_String(0, "te\u0002st");
    });
```

stack trace:
```
System.ArgumentException
'', hexadecimal value 0x02, is an invalid character.
   at System.Xml.XmlUtf8RawTextWriter.InvalidXmlChar(Int32 ch, Byte* pDst, Boolean entitize)
   at System.Xml.XmlUtf8RawTextWriter.WriteElementTextBlock(Char* pSrc, Char* pSrcEnd)
   at System.Xml.XmlUtf8RawTextWriter.WriteString(String text)
   at System.Xml.XmlWellFormedWriter.WriteString(String text)
   at SIL.LCModel.DomainImpl.MultiUnicodeAccessor.ToXml(XmlWriter writer, ILgWritingSystemFactory wsf, Int32 ws, ITsString alternative)
   at SIL.LCModel.DomainImpl.MultiAccessor.ToXMLString(XmlWriter writer)
   at SIL.LCModel.DomainImpl.ReadWriteServices.WriteMultiFoo(XmlWriter writer, String elementName, MultiAccessor multiProperty)
   at SIL.LCModel.DomainImpl.MoForm.ToXMLStringInternal(XmlWriter writer)
   at SIL.LCModel.DomainImpl.MoStemAllomorph.ToXMLStringInternal(XmlWriter writer)
   at SIL.LCModel.DomainImpl.CmObject.SIL.LCModel.ICmObjectInternal.ToXmlBytes()
   at SIL.LCModel.DomainImpl.CmObject.SIL.LCModel.ICmObjectInternal.ToXmlString()
   at SIL.LCModel.Infrastructure.Impl.LcmStateChangeObjectCreation.SetAfterXML()
   at SIL.LCModel.Infrastructure.Impl.UndoableUnitOfWork.SetAfterXml()
[...]

PR #329, opened 2025-06-17
Strip invalid char data from strings on save

hahn-kev:
relates to #328

I didn't see anything stripping out data, so I've gone ahead and done that via a compiled regex.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/329)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- [SIL.LCModel] Add SpecificItemAndFieldName() to ISenseOrEntry
 - [SIL.LCModel] Added a parameter to GetBestGuess() and TryGetBestGuess() to do lowercase matching regardless of the occurrence index
 - [SIL.LCModel] Add GetCaptionOrHeadword() to CmPicture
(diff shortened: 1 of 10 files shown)


hahn-kev:
@jasonleenaylor I've added a link to a Wikipedia page documenting valid xml chars. One thing it mentioned is that `U+10000–U+10FFFF` is also valid, we have not included that here. Is that ok? or do we need to include it? I realized this has the potential to strip valid data on save if we get this wrong. So it might be worth a bit more checking to ensure this is correct otherwise users could start losing data on save without noticing right away.

jasonleenaylor:
Sorry that I didn't see your comment before, yes we should include that range. It is very likely someone using FieldWorks will have a character in that range.

hahn-kev:
@jasonleenaylor I worked with Martin H and rewrote the regex to just match the invalid characters that we want to remove, and he helped me come up with a test case which was outside the normal range.

jasonleenaylor (approved):
> @jasonleenaylor reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/329)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @hahn-kev)

PR #332, opened 2025-07-11
Fix LT-21902: Make MoInflAffixSlot be Cloneable

jtmaxwell3:
liblcm support for https://jira.sil.org/browse/LT-21902.  I make IMoInflAffixSlot cloneable and I make both slots and templates have unique names when they get cloned.

### DIFF
--- a/src/SIL.LCModel/DomainImpl/OverridesLing_MoClasses.cs
+++ b/src/SIL.LCModel/DomainImpl/OverridesLing_MoClasses.cs
@@ -2880,6 +2880,29 @@ namespace SIL.LCModel.DomainImpl
 		}
 
 		#endregion
+
+		#region ICloneableCmObject Members
+		/// ------------------------------------------------------------------------------------
+		/// <summary>
+		/// Clones this instance.
+		/// </summary>
+		/// <param name="clone">Destination object for clone operation</param>
+		/// ------------------------------------------------------------------------------------
+		public void SetCloneProperties(ICmObject clone)
+		{
+			IMoInflAffixSlot slot = clone as IMoInflAffixSlot;
+			if (slot == null)
+				throw new ApplicationException("Failed to copy inflectional affix slot:  the target is not an inflectional affix slote.");
+
+			slot.Name.CopyAlternatives(Name);
+			slot.Description.CopyAlternatives(Description);
+			slot.Optional = Optional;
+			foreach (IMoInflAffMsa affix in Affixes)
+			{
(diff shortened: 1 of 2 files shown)


jasonleenaylor:
*[`src/SIL.LCModel/DomainImpl/OverridesLing_MoClasses.cs` line 2925 at r2](https://reviewable.io/reviews/sillsdev/liblcm/332#-OV9J0XB0fyqUyhHi-zx:-OV9J0XB0fyqUyhHi-zy:bv829uf) ([raw file](https://github.com/sillsdev/liblcm/blob/b6e57e029fc39c30c595c468e67ec2b2196c1fde/src/SIL.LCModel/DomainImpl/OverridesLing_MoClasses.cs#L2925)):*
> ```Smalltalk
> 			foreach (IMoInflAffMsa msa in Affixes)
> 			{
> 				if (msa.Owner is not ILexEntry entry)
> ```

This could use a comment. If there is an item in the affixes which is not an ILexEntry it won't be part of the clone and the reason should be given.

jtmaxwell3:
If it is not a LexEntry, then the next line would crash.  I don't think that it can be anything but a LexEntry, but I'm trying to be more defensive in my programming.

jasonleenaylor:
*[`src/SIL.LCModel/DomainImpl/OverridesLing_MoClasses.cs` line 2925 at r2](https://reviewable.io/reviews/sillsdev/liblcm/332#-OV9J0XB0fyqUyhHi-zx:-OVDcxeOEWx0ttR1WAL8:b-1zcdvf) ([raw file](https://github.com/sillsdev/liblcm/blob/b6e57e029fc39c30c595c468e67ec2b2196c1fde/src/SIL.LCModel/DomainImpl/OverridesLing_MoClasses.cs#L2925)):*

This could use a comment. If there is an item in the affixes which is not an ILexEntry it won't be part of the clone and the reason should be given.
</blockquote></details>

In that case just a comment explaining that this is never expected but we'd rather not crash is good enough.
Just to keep the next developer from wondering.

jasonleenaylor (approved):
:lgtm:

> Reviewed 1 of 2 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/332#-:-OVDojwuEJVM8RUgGMR_:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

PR #334, opened 2025-07-23
Fix LT-21976: Merge Create dates when merging entries

jasonleenaylor:
* Store the date at the start of the import and use it in entry creation. We need to have a consistent date recorded to determine if we read a date from the xml data or not.
* Add a function to return the newest date that was actually read from the xml data.

### DIFF
--- a/src/SIL.LCModel/Application/ApplicationServices/XmlImportData.cs
+++ b/src/SIL.LCModel/Application/ApplicationServices/XmlImportData.cs
@@ -165,6 +165,7 @@ namespace SIL.LCModel.Application.ApplicationServices
 		private ReferenceTracker m_rglinks = new ReferenceTracker();
 		private TextWriter m_wrtrLog;
 		private bool m_createLinks;
+		private DateTime m_importDate = DateTime.Now;
 
 
 		/// ------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/334)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`src/SIL.LCModel/Application/ApplicationServices/XmlImportData.cs` line 491 at r1](https://reviewable.io/reviews/sillsdev/liblcm/334#-OVwgC6d0_rTHCt_Z0l6:-OVwgC6d0_rTHCt_Z0l7:bm19i3f) ([raw file](https://github.com/sillsdev/liblcm/blob/82df3e5a238a6b9f6455c8c2094d5810c6ede8de/src/SIL.LCModel/Application/ApplicationServices/XmlImportData.cs#L491)):*
> ```Smalltalk
> 				return;
> 			}
> 			if (leNew.DateCreated == m_importDate)
> ```

Would it be worth having `>= m_importDate` on these two checks just in case a `.DateCreated` is from `DateTime.Now` on a system with a different clock?

jasonleenaylor:
*[`src/SIL.LCModel/Application/ApplicationServices/XmlImportData.cs` line 491 at r1](https://reviewable.io/reviews/sillsdev/liblcm/334#-OVwgC6d0_rTHCt_Z0l6:-OVxHPsiAfJt3F4wIBs8:b-6ddy42) ([raw file](https://github.com/sillsdev/liblcm/blob/82df3e5a238a6b9f6455c8c2094d5810c6ede8de/src/SIL.LCModel/Application/ApplicationServices/XmlImportData.cs#L491)):*
<details><summary><i>Previously, imnasnainaec (D. Ror.) wrote…</i></summary><blockquote>

Would it be worth having `>= m_importDate` on these two checks just in case a `.DateCreated` is from `DateTime.Now` on a system with a different clock?
</blockquote></details>

No, I don't think it is worth it. The real concern here is retaining data coming in from the import, if we end up with created dates in the future that is a GIGO issue for the user that doesn't cause any real issues for our functionality.

PR #339, opened 2025-08-26
Fix LT-22227: Keep affix order when duplicating an affix slot

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-22227.  It assumes that the order of the old Affixes is stable even if it doesn't have a VirtualOrdering specified.  If that isn't true then the order of the old Affixes should be specified when the order of the new Affixes is specified.  There is no attempt to keep the orders synced after the copy has been made.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/339)

### DIFF
--- a/src/SIL.LCModel/DomainImpl/OverridesLing_MoClasses.cs
+++ b/src/SIL.LCModel/DomainImpl/OverridesLing_MoClasses.cs
@@ -12,6 +12,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using System.Xml;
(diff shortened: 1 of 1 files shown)


AndyBlack (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/339#-:-OYbGuI8-NbEy1k8Ge3t:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved

PR #340, opened 2025-08-27
Fix LT-22155: Merge entry/sense losing some custom fields

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-22155.  The code for custom fields was missing a bunch of types.  I added a "default" case that should catch everything, but I couldn't test all of the possibilities, just the ones in LT-22155.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/340)

### DIFF
--- a/src/SIL.LCModel/DomainImpl/CmObject.cs
+++ b/src/SIL.LCModel/DomainImpl/CmObject.cs
@@ -833,9 +833,16 @@ namespace SIL.LCModel.DomainImpl
 							myCurrentValue = new MultiStringAccessor(this, flid);
 							srcCurrentValue = new MultiStringAccessor(objSrc, flid);
 							break;
-						case (int)CellarPropertyType.MultiUnicode:
-							myCurrentValue = new MultiUnicodeAccessor(this, flid);
-							srcCurrentValue = new MultiUnicodeAccessor(objSrc, flid);
+						case (int)CellarPropertyType.OwningAtomic:
+						case (int)CellarPropertyType.ReferenceAtomic:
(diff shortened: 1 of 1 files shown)


src/SIL.LCModel/DomainImpl/CmObject.cs
```diff
@@ -833,9 +833,16 @@ public void MergeSelectedPropertiesOfObject(ICmObject objSrc, bool fLoseNoString
 							myCurrentValue = new MultiStringAccessor(this, flid);
 							srcCurrentValue = new MultiStringAccessor(objSrc, flid);
 							break;
-						case (int)CellarPropertyType.MultiUnicode:
```
hahn-kev, line 836: it looks like MultiUnicode was removed?

jtmaxwell3: MulitUnicode was removed because it was producing the wrong result.  It now falls through to the default case, which produces the right result.

jasonleenaylor (commented):
> @jasonleenaylor reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/340)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @hahn-kev)

jasonleenaylor (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/340#-:-OYlT4GN6jClrQTXKLyQ:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @hahn-kev)

PR #364, opened 2026-02-09
Rewrite AGENTS.md with architecture guide and add task docs for agents

jasonleenaylor:
Replace the rough-draft AGENTS.md with a comprehensive guide covering critical rules, the code generation pipeline, model schema, persistence infrastructure, and data migration system. Add separate step-by-step task documents in docs/agents/ for common operations: adding properties, writing migrations, adding classes, adding virtual properties, and writing tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/364)


johnml1135:
*[`AGENTS.md` line 61 at r1](https://reviewable.io/reviews/sillsdev/liblcm/364#-Ol39K01D1c_w6iVwffw:-Ol39K01D1c_w6iVwffx:bjvkfxv) ([raw file](https://github.com/sillsdev/liblcm/blob/00c4edaa97c18d115433d622ea096e0fcf2025ac/AGENTS.md#L61)):*
> ```Markdown
> 3. Processes NVelocity templates (`LcmGenerate/*.vm.cs`) to produce the 9 generated C# files
> 
> The generated code provides: class ID/field ID constants, interfaces, concrete implementations, factory interfaces and implementations, repository interfaces and implementations, the backend provider's `ModelVersion` constant, and StructureMap DI bootstrapping.
> ```

Can one more sentence be added as to why this architecture was chosen?  It is unobvious to me.  Is there an article that explains this paradigm?  Do newer versions of DotNet make this irrelevant, but we want to still support mono?  Is this a good long term strategy?  If you feel we need a few paragraphs to explain, can you reference a document or wiki, etc?

johnml1135:
*[`AGENTS.md` line 173 at r1](https://reviewable.io/reviews/sillsdev/liblcm/364#-Ol3AnXVEuhcfsk4VWp6:-Ol3AnXW1tBW-7X7ew2D:brg4y3q) ([raw file](https://github.com/sillsdev/liblcm/blob/00c4edaa97c18d115433d622ea096e0fcf2025ac/AGENTS.md#L173)):*
> ```Markdown
> When performing any of the following tasks, read the linked guide first:
> 
> - **Adding a property to an existing class**: [docs/agents/adding-a-property.md](docs/agents/adding-a-property.md)
> ```

Is it best practice to put these here?  Should they be grouped with the actual relevant data?

johnml1135:
*[`docs/agents/adding-a-property.md` line 2 at r1](https://reviewable.io/reviews/sillsdev/liblcm/364#-Ol3BCZADhW5tEsQdwE7:-Ol3BCZBDwFsyAHahIaE:b2isc6y) ([raw file](https://github.com/sillsdev/liblcm/blob/00c4edaa97c18d115433d622ea096e0fcf2025ac/docs/agents/adding-a-property.md#L2)):*
> ```Markdown
> # Task: Adding a Property to an Existing Class
> ```

Are these files generated from a user guide that was fed into Claude?  How have they been validated?  I don't know enough to know if they are right.

johnml1135:
*[`docs/agents/adding-a-property.md` line 75 at r1](https://reviewable.io/reviews/sillsdev/liblcm/364#-Ol3BZGD5wJPYsfpQONO:-Ol3BZGE8hdnpvxjdeJF:by6n3h4) ([raw file](https://github.com/sillsdev/liblcm/blob/00c4edaa97c18d115433d622ea096e0fcf2025ac/docs/agents/adding-a-property.md#L75)):*
> ```Markdown
> Create: `src/SIL.LCModel/DomainServices/DataMigration/DataMigration7000073.cs`
> 
> **For new optional properties with safe defaults (most common case)**, existing data doesn't need modification. But you still need the migration class:
> ```

Is this a duplication of data migration?  You can ask Claude "Check for duplicated code here"  "Refactor to have an appropriate amount of context for a broad scope of work".

johnml1135:
*[`docs/agents/adding-a-new-class.md` line 3 at r1](https://reviewable.io/reviews/sillsdev/liblcm/364#-Ol3CSkC-q9evZkFSE5J:-Ol3CSkC-q9evZkFSE5K:bfiqnp5) ([raw file](https://github.com/sillsdev/liblcm/blob/00c4edaa97c18d115433d622ea096e0fcf2025ac/docs/agents/adding-a-new-class.md#L3)):*
> ```Markdown
> # Task: Adding a New Class to the Model
> 
> This guide covers adding an entirely new class to `MasterLCModel.xml`. This is less common than adding properties and more involved -- it touches the model, requires a migration, may need an owner property on an existing class, and may need hand-written partial class logic.
> ```

You may also want a impact-assessment subagent (or similar).  Sending off sub-agents to research the codebase and then come back is often effective.  YOu can create a custom agents and then call it out in other agents to use as appropriate.

johnml1135 (commented):
> @johnml1135 reviewed 6 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/364)* status: all files reviewed, 15 unresolved discussions (waiting on @jasonleenaylor).

jasonleenaylor (commented):
> @jasonleenaylor made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/364)* status: all files reviewed, 15 unresolved discussions (waiting on @johnml1135).

___
*[`AGENTS.md` line 21 at r1](https://reviewable.io/reviews/sillsdev/liblcm/364#-Ol37viL9j8axgHJTlcg:-Ol3IvVg0ZOG3-1mE1As:b-ehcmom) ([raw file](https://github.com/sillsdev/liblcm/blob/00c4edaa97c18d115433d622ea096e0fcf2025ac/AGENTS.md#L21)):*
<details><summary><i>Previously, johnml1135 (John Lambert) wrote…</i></summary><blockquote>

Can this be enforced?  Can the write commands only be accessible through these two commands?
</blockquote></details>

It is enforced actually, this is just making sure the agent understands that. Actually there is better guidance I can give here. The silently fail is not accurate.
___
*[`AGENTS.md` line 23 at r1](https://reviewable.io/reviews/sillsdev/liblcm/364#-Ol383MZ2qhIHEiTRdxf:-Ol3JDq45C9xDYQ3JyaB:b-al7cky) ([raw file](https://github.com/sillsdev/liblcm/blob/00c4edaa97c18d115433d622ea096e0fcf2025ac/AGENTS.md#L23)):*
<details><summary><i>Previously, johnml1135 (John Lambert) wrote…</i></summary><blockquote>

Why would Windows Forms be in here at all?  Why reference it if it is not in the repo?  Does this need one more sentence or clause of explanation?
</blockquote></details>

We specifically guard against developers who tend to add things like message boxes or other things from System.Windows.Forms (there is a build target that guards against it). The AI picked up on that, which seems fine.
___
*[`AGENTS.md` line 48 at r1](https://reviewable.io/reviews/sillsdev/liblcm/364#-Ol38lgl-659gfQFqIVi:-Ol3JedXB423Fp0YRImE:b-w00slj) ([raw file](https://github.com/sillsdev/liblcm/blob/00c4edaa97c18d115433d622ea096e0fcf2025ac/AGENTS.md#L48)):*
<details><summary><i>Previously, johnml1135 (John Lambert) wrote…</i></summary><blockquote>

Can we fix this programmatically?  When we start a build (and need it), can we ensure that we are not in a shallow clone in a quick way?  A fix is better than a known issue...
</blockquote></details>

It is a known GitVersion issue I believe, not an issue of ours.
___
*[`AGENTS.md` line 61 at r1](https://reviewable.io/reviews/sillsdev/liblcm/364#-Ol39K01D1c_w6iVwffw:-Ol3KXtn0yh0YviaFWVJ:be5tb80) ([raw file](https://github.com/sillsdev/liblcm/blob/00c4edaa97c18d115433d622ea096e0fcf2025ac/AGENTS.md#L61)):*
<details><summary><i>Previously, johnml1135 (John Lambert) wrote…</i></summary><blockquote>

Can one more sentence be added as to why this architecture was chosen?  It is unobvious to me.  Is there an article that explains this paradigm?  Do newer versions of DotNet make this irrelevant, but we want to still support mono?  Is this a good long term strategy?  If you feel we need a few paragraphs to explain, can you reference a document or wiki, etc?
</blockquote></details>

If you are referring to the class id, field id constants that was to provide fast reflection and support data migration capabilities which didn't require instantiating C# classes. It has nothing to do with mono. 
This markdown is specific to the agents, so I don't know that adding paragraphs to explain it is valuable.
___
*[`AGENTS.md` line 173 at r1](https://reviewable.io/reviews/sillsdev/liblcm/364#-Ol3AnXVEuhcfsk4VWp6:-Ol3JsOa2EEL3I8LqC8i:b-uiuesx) ([raw file](https://github.com/sillsdev/liblcm/blob/00c4edaa97c18d115433d622ea096e0fcf2025ac/AGENTS.md#L173)):*
<details><summary><i>Previously, johnml1135 (John Lambert) wrote…</i></summary><blockquote>

Is it best practice to put these here?  Should they be grouped with the actual relevant data?
</blockquote></details>

I'm not sure if this is best practice. this is what I got when I asked it to keep the context which an agent starts with as small as reasonable but to have instructions for specific common tasks with best practices. I was wondering about this and thinking they should really be 'skills' that live somewhere. I'm not sure on the best practice if there is one for vendor unlocked skills files. 
___
*[`docs/agents/adding-a-property.md` line 2 at r1](https://reviewable.io/reviews/sillsdev/liblcm/364#-Ol3BCZADhW5tEsQdwE7:-Ol3MlNMAWU8eukbgMd2:b-sfvk21) ([raw file](https://github.com/sillsdev/liblcm/blob/00c4edaa97c18d115433d622ea096e0fcf2025ac/docs/agents/adding-a-property.md#L2)):*
<details><summary><i>Previously, johnml1135 (John Lambert) wrote…</i></summary><blockquote>

Are these files generated from a user guide that was fed into Claude?  How have they been validated?  I don't know enough to know if they are right.
</blockquote></details>

Yes, these files came from a combination of a model description pdf and specific code file guidance that I gave. I've mostly validated them but I was actually executing a test model change to validate some of this.
___
*[`docs/agents/writing-a-data-migration.md` line 3 at r1](https://reviewable.io/reviews/sillsdev/liblcm/364#-Ol3CnHz-eTWVRKXMrxS:-Ol3NS0UC7vU0LkPU0dG:b-lrk0cj) ([raw file](https://github.com/sillsdev/liblcm/blob/00c4edaa97c18d115433d622ea096e0fcf2025ac/docs/agents/writing-a-data-migration.md#L3)):*
<details><summary><i>Previously, johnml1135 (John Lambert) wrote…</i></summary><blockquote>

This sounds very formal.  Will this become the formal documentation or is it written somewhere else?  Can we reference it or can this become it?
</blockquote></details>

This is a description of how the datamigration code works and it is giving the agent instructions on how to successfully add one (I'm currently testing if these instructions actually result in successful agent action)

johnml1135:
*[`tests/SIL.LCModel.Core.Tests/Attributes/InitializeIcuAttribute.cs` line 6 at r2](https://reviewable.io/reviews/sillsdev/liblcm/364#-OlImStR18fXnSsiz3Jp:-OlImouqDrnDpaifxHR9:b-rm4dwl) ([raw file](https://github.com/sillsdev/liblcm/blob/a06c52fe74fa8bddfee46256c5e629d0bd516796/tests/SIL.LCModel.Core.Tests/Attributes/InitializeIcuAttribute.cs#L6)):*
<details><summary><i>Previously, johnml1135 (John Lambert) wrote…</i></summary><blockquote>

What are these other changes for?  the removal of the proj file?  I am assuming you moved from the script to dotnet build.  Is that all?  Can you add a comment for these changes?
</blockquote></details>

I see the git comment - does that match your understanding as to what happened?

hahn-kev:
this is super helpful for human agents as well 😄

jasonleenaylor:
@johnml1135 — answering your two comments on `InitializeIcuAttribute.cs` at r2. Sorry for the long delay.

Good questions, and the answer turned out to be "no, that wasn't all" — thanks for pushing on it.

Three things were happening in that file, not one. The new `SetIcuEnvironmentFromBuildOutputIfPresent` helper set `ICU_DATA` and prepended `lib` paths; `Wrapper.Init()` was removed; and `ConfineIcuVersions` moved after initialization. Only the second one was load-bearing.

`EnsureIcuDataEnvironmentVariableIsSet` already resolves the same directory, because every test assembly declares `[assembly: InitializeIcu(IcuDataPath = "IcuData")]` and that takes the relative-path branch against the build output. And `CustomIcu.InitIcuDataDir` already prepends the same four `lib` directories. So the helper was doing nothing that wasn't done immediately after it — I've deleted it, and the reordering with it.

The actual defect: `Wrapper.Init()` ran before `PATH` had the `lib` directories and before `ICU_DATA` was set, so the first native ICU load had nothing to find. That's what made agents set the environment by hand. Removing that call is the whole fix — the diff against master is now +2/-10, and I've corrected the catch message that still said `Wrapper.Init()` while guarding `InitIcuDataDir()`.

Verified with `ICU_DATA` unset, no ICU entries on `PATH`, no `Software\SIL\Icu70DataDir` registry value and no `%ProgramData%\SIL\Icu70`: `SIL.LCModel.Core.Tests` 792/792, `SIL.LCModel.FixData.Tests` 21/21 and `SIL.LCModel.Tests` 1716/1716, all on net462 and net8.0.

On the commit message — the third bullet was accurate, but it under-described the file, and the subject named the AGENTS.md trim when most of the commit was deleting `build.cmd`, `build.sh` and `build/LCM.proj`. That's fair criticism.

So I've split this PR. This one now keeps only the build-script removal, the ICU fix, and one `global.json` change; it's rebased onto current master and retitled. The AGENTS.md rewrite and the task skills will come separately, rewritten against today's tree rather than February's.

The `global.json` change is new since you last looked, so flagging it explicitly: `rollForward` goes from `latestFeature` to `latestMajor`. The `8.0.100` pin added in #394 refuses any newer SDK, which blocks work on a machine with no 8.0.x SDK installed. CI installs 8.0.x explicitly via `setup-dotnet`, so the pin was redundant there. Verified the solution builds clean and the suites pass on both 9.0.316 and 10.0.201.

PR #372, opened 2026-05-01
Fix LT-21889: Add Find and Fix step for removing disowned children

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-21889.  If the owner and stored owner don't match and the owner exists and the stored owner doesn't exist then the node has been disowned.  This revealed some problems with the unit tests (type="o" instead of t="o").

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/372)

### DIFF
--- a/src/SIL.LCModel.FixData/FwDataFixer.cs
+++ b/src/SIL.LCModel.FixData/FwDataFixer.cs
@@ -256,6 +256,11 @@ namespace SIL.LCModel.FixData
 							errorLogger(String.Format(Strings.ksRemovingObjectWithBadOwner, guidOwner, className, guid), true);
 							return false;
 						}
+						else
+						{
+							errorLogger(String.Format(Strings.ksRemovingDisownedObject, guidOwner, className, guid), true);
+							return false;
+						}
 					}
 				}
 				else if (storedOwner != Guid.Empty)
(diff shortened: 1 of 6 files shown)


jasonleenaylor:
*[`tests/SIL.LCModel.FixData.Tests/FwDataFixerTests.cs` line 484 at r1](https://reviewable.io/reviews/sillsdev/liblcm/372#-Or_GWNzCmOJyZrlBYCx:-Or_GWO-BVCvHGZVFJJz:b-se0jd2) ([raw file](https://github.com/sillsdev/liblcm/blob/f355c09a7a86d62a91f7c507ea4a8ff151d1e1d0/tests/SIL.LCModel.FixData.Tests/FwDataFixerTests.cs#L484)):*
> ```Smalltalk
> 			Assert.That(_errors[8], Is.EqualTo("Removing dangling link to Form '" + danglingMsaGuid + "' for WfiMorphBundle '" + danglingMorphNoRepairGuid + "'."),
> 				"Error message is incorrect."); // MorphBundleFixer--ksRemovingDanglingMorph
> 			Assert.True(_errors[9].StartsWith("Removing disowned object (invalid ownerguid='" + disownedOwnerGuid),
> ```

The added assert for the error message is good, but we should also have a new assert for the actual data being removed. I expected to see an assert below in the `// Check file repair section `

jasonleenaylor (approved):
:lgtm:

> @jasonleenaylor partially reviewed 6 files and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/372#-:-OrnYSjAATgM2V0c65gy:bnfp4nl)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on jtmaxwell3).

PR #375, opened 2026-05-13
Expose CorrectHomographNumbers publicly on ILexEntry

myieye:
Adds `ILexEntry.CorrectHomographNumbers()` which wraps the existing `LexDb.CorrectHomographNumbers` algorithm.
FieldWorks Lite writes `HomographNumber` directly and needs a way to trigger a "cleanup" of a potentially broken homograph sequence.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/375)

### DIFF
--- a/src/SIL.LCModel/DomainImpl/OverridesLing_Lex.cs
+++ b/src/SIL.LCModel/DomainImpl/OverridesLing_Lex.cs
@@ -2044,6 +2044,20 @@ namespace SIL.LCModel.DomainImpl
 			return result;
 		}
 
+		/// <summary>See <see cref="ILexEntry.CorrectHomographNumbers"/>.</summary>
+		public bool CorrectHomographNumbers()
+		{
+			if (HomographFormKey == Strings.ksQuestions)
+			{
+				if (HomographNumber == 0) return true;
+				HomographNumber = 0;
+				return false;
+			}
+			var homographs = Services.GetInstance<ILexEntryRepository>()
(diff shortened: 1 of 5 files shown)


jasonleenaylor (changes requested):
From an API standpoint I don't like adding a method on LexEntry which modifies data in other LexEntries. This should be exposed, but I don't think it belongs here.

> @jasonleenaylor reviewed all commit messages and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/375#-:-OsXBZIF7EplPAt698oR:b-j0r6f5)* status: 0 of 3 files reviewed, all discussions resolved.

myieye:
@jasonleenaylor You're right. I moved it.

API-wise it would be arguably better to make the signature of my new method more similar to `CollectHomographs` in that it just takes a `string form` and morph-type. But it would be a bit awkward if it couldn't be used to correct blank-form entries. To do that it obviously needs the entry itself. If you're unhappy with it taking an entry and then **also** editing **other** entries. Then I can change it and make FW Lite handle the blank-form case. The current code would just be a bit more handy for us.

jasonleenaylor (approved):
:lgtm:

> @jasonleenaylor reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/375#-:-Ot-wMIs1DSjWyJvSCNO:bnfp4nl)* status: 4 of 5 files reviewed, all discussions resolved.

PR #378, opened 2026-05-26
Fix LT-18897: FLEx is silently changing completed analyses

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-18897.  AnalysisAdjuster wasn't recognizing that a segment basically hadn't changed if it included some extra white space.  I added DeltaIsEmpty to catch this case.  So, adding a paragraph break anywhere between two segments should work now.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/378)

### DIFF
--- a/src/SIL.LCModel/DomainImpl/AnalysisAdjuster.cs
+++ b/src/SIL.LCModel/DomainImpl/AnalysisAdjuster.cs
@@ -1478,10 +1478,6 @@ namespace SIL.LCModel.DomainImpl
 		/// Is the delta between the old segment boundary and the new segment boundary empty?
 		/// This fixes LT-18897.
 		/// </summary>
-		/// <param name="endOffset"></param>
-		/// <param name="ichSource"></param>
-		/// <param name="source"></param>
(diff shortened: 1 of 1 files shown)


src/SIL.LCModel/DomainImpl/AnalysisAdjuster.cs
```diff
@@ -1474,6 +1474,28 @@ private void DetermineWhatToMove(int ichSource, int cchInsert, IStTxtPara source
...
 		}
 
+		/// <summary>
+		/// Is the delta between the old segment boundary and the new segment boundary empty?
+		/// This fixes LT-18897.
+		/// </summary>
+		/// <param name="endOffset"></param>
+		/// <param name="ichSource"></param>
+		/// <param name="source"></param>
+		/// <returns></returns>
```
mark-sil: Probably should not include the comments for the params and the return if they don't contain any information.

mark-sil (approved):

PR #380, opened 2026-06-18
Fix LT-21598: Crash deleting from Bulk Edit Entries

mark-sil:
## Summary
Bulk deleting lexical entries that participate in a lexical relation could crash with a `NullReferenceException` ([LT-21598](https://jira.sil.org/browse/LT-21598)). The attached project (Puntipi-Luritja) has corrupt data — a lost incoming-reference index, not repaired by Find and Fix — so a `LexReference` can still list a target that has already been deleted. During the deletion cascade, `LexReference`'s removal side effects operated on those already-deleted objects and dereferenced their cleared cache.

## Changes (`LexReference` in `OverridesLing_Lex.cs`)
- `UpdateMinimalLexReferences` skips targets that are no longer valid.
- `RemoveObjectSideEffectsInternal` only updates the removed object's back-reference virtual properties (`LexEntryReferences` / `LexSenseReferences` / `DateModified`) while it is still valid; the remaining valid targets are still refreshed via `UpdateMinimalLexReferences`.

## Testing
- Added two regression tests in `LingTests.cs`, each verified to fail at the exact reported stack without its fix.
- Full LCM test suite passes (net8.0): **2806 passed, 0 failed, 20 skipped**.
- Verified manually in FLEx: Bulk Edit Entries → Delete on the Puntipi-Luritja project no longer crashes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/380)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- [SIL.LCModel] Fixed crash when bulk deleting entries involved in a lexical relation (LT-21598)
 - [SIL.LCModel] Data migration now serializes dates using the culture-neutral `ToLCMTimeFormatWithMillisString` (LT-20698)
 - [SIL.LCModel] `ReadWriteServices.LoadDateTime` now parses milliseconds correctly (LT-18205)
 - [SIL.LCModel.Core] Copy `SIL.LCModel.Core.dll.config` to output directory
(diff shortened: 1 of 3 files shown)


hahn-kev (commented):
Looks good to me, the tests reproduce the issue and fail without the fix.

One thing I did notice, that may effect something, is that the order in which `UpdateMinimalLexReferences` and `UpdateLexEntryReferences` is called has changed. The old version would call `UpdateLex` first, then `UpdateMinimal` however the new version changes that. I'm not sure if this was intentional but we might consider putting it back to how it was.

mark-sil (commented):
Thanks Kevin, I restored the order.

> @mark-sil made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/380#-:-OvoU8XD-MhXzugO6lph:beqw8nz)* status: 0 of 3 files reviewed, all discussions resolved.

hahn-kev (approved):
looks good to me, I don't have the full context of this and I'm not sure about the original issue though.

PR #383, opened 2026-07-01
LT-14056: Crash with Undo then Redo deleted Lexical Relations

aror92:
Add null cache guards. The redo is deleting a lexical relation that was just "undeleted" via undo. Tries to restore refs when the object has already been deleted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/383)

### DIFF
--- a/src/SIL.LCModel/DomainImpl/CmObject.cs
+++ b/src/SIL.LCModel/DomainImpl/CmObject.cs
@@ -2629,11 +2629,8 @@ namespace SIL.LCModel.DomainImpl
 		{
 			// Add a null cache guard. (See LT-14056: when using redo button to delete a lexical relation that was "undeleted" via undo,
 			// it tries to restore refs after deleting the object and crashes b/c Cache is null.)
-			if (Cache == null)
-			{
-				Debug.Fail("Restoring incoming refs on outgoing refs for an object already deleted");
-				return;
-			}
+			if (Cache == null) { return; }
+
 			var mdc = Cache.ServiceLocator.GetInstance<IFwMetaDataCacheManaged>();
 			foreach (var flid in mdc.GetFields(ClassID, true, (int)CellarPropertyTypeFilter.AllReference))
 			{
--- a/src/SIL.LCModel/DomainImpl/CmObject.cs
+++ b/src/SIL.LCModel/DomainImpl/CmObject.cs
@@ -2627,6 +2627,13 @@ namespace SIL.LCModel.DomainImpl
 		/// </summary>
 		protected virtual void RestoreIncomingRefsOnOutgoingRefsInternal()
 		{
+			// Add a null cache guard. (See LT-14056: when using redo button to delete a lexical relation that was "undeleted" via undo,
+			// it tries to restore refs after deleting the object and crashes b/c Cache is null.)
+			if (Cache == null)
+			{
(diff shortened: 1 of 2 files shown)


mark-sil (changes requested):
> @mark-sil reviewed all commit messages and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/383)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on aror92).

___
*[`src/SIL.LCModel/DomainImpl/CmObject.cs` line 2634 at r1](https://reviewable.io/reviews/sillsdev/liblcm/383#-OwT790E489GYmdkjSpP:-OwT790F3cz9PgEKJqyf:b-6qy53l) ([raw file](https://github.com/sillsdev/liblcm/blob/65ac4fa1cc00e0c1e881cf536e43991420a03a82/src/SIL.LCModel/DomainImpl/CmObject.cs#L2634)):*
> ```Smalltalk
> 			if (Cache == null)
> 			{
> 				Debug.Fail("Restoring incoming refs on outgoing refs for an object already deleted");
> ```

Why do we want the Debug.Fail() messages (both this one and the other)? To me they imply that there is something wrong that needs to be investigated. If the correct solution is to just return (as you are doing), then we probably shouldn't be alerting the developers.

aror92 (commented):
> @aror92 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/383)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on mark-sil).

___
*[`src/SIL.LCModel/DomainImpl/CmObject.cs` line 2634 at r1](https://reviewable.io/reviews/sillsdev/liblcm/383#-OwT790E489GYmdkjSpP:-OwTAKWhBoOfAAMQc8s3:bysusma) ([raw file](https://github.com/sillsdev/liblcm/blob/65ac4fa1cc00e0c1e881cf536e43991420a03a82/src/SIL.LCModel/DomainImpl/CmObject.cs#L2634)):*
<details><summary><i>Previously, mark-sil (Mark Kidder) wrote…</i></summary><blockquote>

Why do we want the Debug.Fail() messages (both this one and the other)? To me they imply that there is something wrong that needs to be investigated. If the correct solution is to just return (as you are doing), then we probably shouldn't be alerting the developers.
</blockquote></details>

The Debug.Fail was an addition recommended by Devin/LcmDeepWiki. I was torn on it, but Devin pointed out that we already have one guard for the same reason in ClearIncomingRefsOnOutgoingRefsInternal, and in that case, it includes the comment "SHOULD never be called on already deleted object" and uses a Debug.Fail message for that reason.

The same logic would apply to both of these guards, so I included the Debug.Fail messages. (i.e. Cache == null means undo/redo is operating on an already deleted object, which we appear to claim should never happen by design.) I'm okay removing the Debug.Fails, though, if we're sure we don't want them.

mark-sil (approved):
> @mark-sil reviewed 2 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/383)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on aror92).

PR #384, opened 2026-07-02
Replace System.Web with System.Net.WebUtility in data migration

hahn-kev:
This allows us to remove our dependency on `System.Web.HttpUtility` which is designed for use on the server per https://learn.microsoft.com/en-us/dotnet/api/system.web.httputility?view=net-10.0#remarks 

for DataMigration7000040Tests.cs I removed the unused copy-pasted helpers that still referenced System.Web and were never called by the test.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/384)

### DIFF
--- a/src/SIL.LCModel/DomainServices/DataMigration/DataMigration7000037.cs
+++ b/src/SIL.LCModel/DomainServices/DataMigration/DataMigration7000037.cs
@@ -7,8 +7,8 @@
 
 using System;
 using System.IO;
+using System.Net;
 using System.Text;
-using System.Web;
 using System.Xml.Linq;
 using System.Xml.XPath;
 using SIL.LCModel.Utils;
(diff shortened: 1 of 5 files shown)


imnasnainaec (approved):
> @imnasnainaec reviewed 5 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/384)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on hahn-kev).

___
*[a discussion](https://reviewable.io/reviews/sillsdev/liblcm/384#-OwY73se84vdc602Bhy8:-OwY73sf5Hw1Ig2TV4a4:bfvgy56) (no related file):*
One observation by Devin:

> src/SIL.LCModel/DomainServices/DataMigration/DataMigration7000037.cs:133
> 
> **URL encoding output will have different hex casing in migrated data**
> 
> The switch from `HttpUtility.UrlEncode` to `WebUtility.UrlEncode` (`DataMigration7000037.cs:133`) produces uppercase hex digits (`%3D`, `%3A`, `%26`) instead of lowercase (`%3d`, `%3a`, `%26`). For example, `app=flex&tool=interlinearEdit` would previously encode as `app%3dflex%26tool%3dinterlinearEdit` but now encodes as `app%3Dflex%26tool%3DinterlinearEdit`. Both are valid per RFC 3986 and decode identically, but any database that still needs this migration applied will get cosmetically different stored data than it would have under the old code. Since this is a one-time migration (version 7000036→7000037) and most databases have long since been migrated, the practical impact is negligible.

PR #395, opened 2026-08-14
LT-22524: Add substring search mode to StringSearcher

thejambi:
Add a Substring value to SearchType that matches the query anywhere within a string, case- and diacritic-insensitive, backed by a raw-string index scanned with CompareInfo.IndexOf. The existing Exact/Prefix/FullText modes are unchanged. This change enables a change in FieldWorks to use this new SearchType, see https://github.com/sillsdev/FieldWorks/pull/1069

https://jira.sil.org/browse/LT-22524

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/395)


src/SIL.LCModel.Core/Text/StringSearcher.cs
```diff
@@ -268,6 +278,16 @@ public IEnumerable<T> Search(int indexId, int wsId, string text)
...
 					return results;
+
+				case SearchType.Substring:
+					{
+						List<KeyValuePair<T, string>> raw;
+						if (!m_rawIndices.TryGetValue(Tuple.Create(indexId, wsId), out raw))
+							return Enumerable.Empty<T>();
+						CompareInfo ci = CultureInfo.InvariantCulture.CompareInfo;
+						return raw.Where(kv => ci.IndexOf(kv.Value, text,
+							CompareOptions.IgnoreCase | CompareOptions.IgnoreNonSpace) >= 0).Select(kv => kv.Key);
```
hahn-kev: we might need to consider if this will work correctly for us, or if we need to create a manged wrapper in icu-dotnet for string search so we can use what ICU provides for this use case.

src/SIL.LCModel.Core/Text/StringSearcher.cs
```diff
@@ -120,6 +125,7 @@ public IEnumerable<T> GetItems(byte[] lower, byte[] upper)
 		#endregion SortKeyIndex class
 
 		private readonly Dictionary<Tuple<int, int>, SortKeyIndex> m_indices = new Dictionary<Tuple<int, int>, SortKeyIndex>();
+		private readonly Dictionary<Tuple<int, int>, List<KeyValuePair<T, string>>> m_rawIndices = new Dictionary<Tuple<int, int>, List<KeyValuePair<T, string>>>();
```
hahn-kev: I wish I had a suggestion, but this type is really ugly lol. I might replace KeyValuePair with a custom struct which does the same thing but is more understandable what's going on when you look at it.

Another idea (bigger tweak) is to create a custom struct (maybe named SearchBucket or IndexKey) which is a new struct that replaces all the uses of `Tuple<int, int>` (which is quite meaning less IMO), you would just need to match the compare and equality rules of the Tuple. This more falls in line with leaving code better than when you found it.

src/SIL.LCModel.Core/Text/StringSearcher.cs
hahn-kev: In the new SearchType we're now getting the index and never using it, it also is creating the index if it doesn't exist which isn't great since it's never used. The same issue exists on the Search method.

src/SIL.LCModel.Core/Text/StringSearcher.cs
```diff
@@ -195,6 +201,10 @@ public void Add(T item, int indexId, int wsId, string text)
 					foreach (string token in RemoveWhitespaceAndPunctTokens(m_tokenizer(wsId, text)))
 						index.Add(m_sortKeySelector(wsId, token), item);
 					break;
+
+				case SearchType.Substring:
+					GetRawIndex(indexId, wsId).Add(new KeyValuePair<T, string>(item, text ?? string.Empty));
```
hahn-kev: we're handling the case of a null text argument here by using an empty string, but searching for empty strings is explicitly banned. Should we just not add a value to the index if the text is null or empty? We could probably safely do this on the first line of this method. Taking a quick look at the `EntryGoSearchEngine` and others it looks like they never call `Add` with an empty string so that's probably safe to change.

hahn-kev (approved):
This looks good. I left some suggestions, nothing blocking.

One thing, since the issue specifically calls out matching how FW Lite search works, here's some of our tests which would be relevant here
https://github.com/sillsdev/languageforge-lexbox/blob/bbcc058f267a789302b9570e9542d30968199a3f/backend/FwLite/MiniLcm.Tests/QueryEntryTestsBase.cs#L412-L455

Up to you how closely you try to match what FW Lite does.

thejambi:
Thank you! I'm looking into getting some of the enhancements you suggested in.

src/SIL.LCModel.Core/Text/StringSearcher.cs
thejambi: Created a SubstringEntry struct to help clean this up a bit.

src/SIL.LCModel.Core/Text/StringSearcher.cs
thejambi: Done.

src/SIL.LCModel.Core/Text/StringSearcher.cs
thejambi: It seems that this is the same behavior as FWLite's search, so hopefully it will be a good way to go?

I did find another behavior to include, see the "Fold diacritics only when the search term has none" change here.

thejambi (commented):
> @thejambi made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/395)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on hahn-kev).

src/SIL.LCModel.Core/Text/StringSearcher.cs
```diff
@@ -124,8 +124,31 @@ public IEnumerable<T> GetItems(byte[] lower, byte[] upper)
...
+			private readonly string m_text;
+
+			public SubstringEntry(T item, string text)
+			{
+				m_item = item;
+				m_text = text;
+			}
+
+			public T Item { get { return m_item; } }
+			public string Text { get { return m_text; } }
```
hahn-kev, line 145: you could simplify this to `public string Text => m_text;`

hahn-kev (approved):
Nice! Looks great, I just left one minor suggestion, feel free to ignore it.

thejambi (commented):
> @thejambi reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/395)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on hahn-kev).

PR #396, opened 2026-08-14
Fix Windows CI by scoping IntermediateOutputPath per framework

thejambi:
Directory.Build.props sets AppendTargetFrameworkToOutputPath=false so that OutputPath can place output under artifacts/$(Configuration)/$(TargetFramework) itself. That flag also strips $(TargetFramework) from IntermediateOutputPath, so all three target frameworks of a multi-targeted project shared a single obj folder - including the reference assembly under obj/ref. Whichever framework built last owned it.

ProjectReference resolution reads that reference assembly, so the netstandard2.0 compile of SIL.LCModel could be handed the net8.0 build of SIL.LCModel.Core:

  CSC : error CS1705: Assembly 'SIL.LCModel.Core' ... uses
  'System.Drawing.Primitives, Version=8.0.0.0' which has a higher version than
  referenced assembly 'System.Drawing.Primitives, Version=4.0.2.0'
  [SIL.LCModel.csproj::TargetFramework=netstandard2.0]

Build ordering decided whether this hit, which is why Windows failed while Linux stayed green. #394 removed the OutDir override in SIL.LCModel and fixed the bin side of the same problem; the intermediate side kept sharing one path.

Set IntermediateOutputPath to obj/$(Configuration)/$(TargetFramework)/ in Directory.Build.props, and drop the obj/x86/$(Configuration)/ override in SIL.LCModel.Core.csproj that predates multi-targeting and now conflicts with it - Platform is Any CPU repo-wide, so the x86 segment bought nothing. OutputPath and the artifacts layout are unchanged.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/396)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -14,13 +14,6 @@
     <Configuration Condition="'$(Configuration)' == ''">Debug</Configuration>
     <OutputPath>$(MSBuildThisFileDirectory)artifacts/$(Configuration)/$(TargetFramework)</OutputPath>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
-    <!-- AppendTargetFrameworkToOutputPath=false strips $(TargetFramework) from
-         IntermediateOutputPath as well as OutputPath, so every target framework of a
-         multi-targeted project would share one obj folder - and therefore one reference
-         assembly in obj/ref. Whichever framework built last won, which is how a
-         netstandard2.0 compile ended up referencing the net8.0 build of SIL.LCModel.Core
-         (CS1705, System.Drawing.Primitives 8.0.0.0). OutputPath already carries the
-         framework, so only the intermediate path needs restoring here. -->
     <IntermediateOutputPath Condition="'$(TargetFramework)' != ''">obj/$(Configuration)/$(TargetFramework)/</IntermediateOutputPath>
     <PackageOutputPath>$(MSBuildThisFileDirectory)/artifacts</PackageOutputPath>
     <WarningsAsErrors>NU1605;CS8002</WarningsAsErrors>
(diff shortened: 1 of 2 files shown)


jasonleenaylor:
*[`src/SIL.LCModel.Core/SIL.LCModel.Core.csproj` line 89 at r1](https://reviewable.io/reviews/sillsdev/liblcm/396#-P-0JEUYA7z03TVRgmtb:-P-0JEUYA7z03TVRgmtc:b-15n3vd) ([raw file](https://github.com/sillsdev/liblcm/blob/5977839b7e39615a8f5e894dc1c8baf1519260e4/src/SIL.LCModel.Core/SIL.LCModel.Core.csproj#L89)):*
> ```XML
> 
>   <PropertyGroup>
>

thejambi:
*[`src/SIL.LCModel.Core/SIL.LCModel.Core.csproj` line 89 at r1](https://reviewable.io/reviews/sillsdev/liblcm/396#-P-0JEUYA7z03TVRgmtb:-P-0Wt0720DTNy7i5hxk:bmzslbl) ([raw file](https://github.com/sillsdev/liblcm/blob/5977839b7e39615a8f5e894dc1c8baf1519260e4/src/SIL.LCModel.Core/SIL.LCModel.Core.csproj#L89)):*

This comment is unnecessary, no one will come to this file and go 'huh, why is this property that I don't know about mssing?'
</blockquote></details>

Comments have been removed.

jasonleenaylor (approved):
:lgtm:

> @jasonleenaylor reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/liblcm/396#-:-P-0awM82YbbzjOY7uMd:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on thejambi).

